PR #3, opened 2016-07-05
fix version number

bparees:


bparees:
@hhorak ptal

hhorak:
We don't have automatic tests working so far, but this is just a change in comment, so merging.

PR #5, opened 2016-07-12
Added cccp.yml files for the CentOS Container Pipeline

mohammedzee1000:
This is for building these containers through the CentOS
pipeline.


hhorak:
This PR does not influence building, so merging.

PR #9, opened 2016-08-09
Fix squashing in CentOS

omron93:
Docker-squash requires docker_py >= 1.7.2, so "easy_install -q --user docker_py==1.6.0 docker-squash==1.0.0rc6" resolves to install latest docker_py version which requires new setuptools with environment markers support. So this command success in CentOS only with latest rh-python35.

Also updated to use latest docker-squash.

@hhorak @bparees Please test. I will create same PR for other repositories after this PR is complete and merged. Thanks.

### DIFF
--- a/hack/build.sh
+++ b/hack/build.sh
@@ -48,7 +48,7 @@ function docker_build_with_version {
 function squash {
   # FIXME: We have to use the exact versions here to avoid Docker client
   #        compatibility issues
-  easy_install -q --user docker_py==1.6.0 docker-squash==1.0.0rc6
+  easy_install -q --user docker_py==1.7.2 docker-squash==1.0.1
   base=$(awk '/^FROM/{print $2}' $1)
   ${HOME}/.local/bin/docker-squash -f $base ${IMAGE_NAME}
 }


hhorak:
@omron93 Not sure I understand -- does it mean that keeping `docker_py==1.6.0 docker-squash==1.0.0rc6` would require rh-python35, but using `docker_py==1.7.2 docker-squash==1.0.1` does work with system python as well?

omron93:
In CentOS for example with system setuptools (python-setuptools package), command `easy_install --user docker_py==1.6.0 docker-squash==1.0.0rc6` fails with this error

```
Installed /root/.local/lib/python2.7/site-packages/docker_squash-1.0.0rc6-py2.7.egg
Processing dependencies for docker-squash==1.0.0rc6
Searching for docker-py>=1.7.2
Best match: docker-py 1.9.0
Downloading https://pypi.python.org/packages/d7/32/1e6be8c9ebc7d02fe74cb1a050008bc9d7e2eb9219f5d5e257648166e275/docker-py-1.9.0.tar.gz#md5=5d0b1736cbe98f3a157acd14a61ba2ef
Processing docker-py-1.9.0.tar.gz
Writing /tmp/easy_install-efAPgt/docker-py-1.9.0/setup.cfg
Running docker-py-1.9.0/setup.py -q bdist_egg --dist-dir /tmp/easy_install-efAPgt/docker-py-1.9.0/egg-dist-tmp-Udx6NO
error: Setup script exited with error in docker-py setup command: Invalid environment marker: python_version < "3.5"
```

It success only with setuptools from rh-python35 (rh-python35-python-setuptools - this is a newer version which supports 'environment markers') and enabled rh-python35 collection.

Simillar to docker_py issue - https://github.com/docker/docker-py/issues/1019

> but using docker_py==1.7.2 docker-squash==1.0.1 does work with system python as well?

Yes it works with both.

hhorak:
Thanks for explanation, it makes sense then of course. LGTM if the test passes.

omron93:
Openshift jenkins PR testing is using SKIP_SQUASH=1. So this failure was not found...

@hhorak Official RH images are also squashed, or not?

hhorak:
> Openshift jenkins PR testing is using SKIP_SQUASH=1. So this failure was not found...

Ah, haven't realize this, so I'll test manually..

> @hhorak Official RH images are also squashed, or not?

Yes, but the squashing is done by OSBS, so this script is not involved there.

omron93:
> Ah, haven't realize this, so I'll test manually..

For non Openshift images it is testing with squashing enable...

Tests are passing, but I don't know why jenkins is spawning new jobs again and again.

> Yes, but the squashing is done by OSBS, so this script is not involved there.

So should we squash images during PR testing?

bparees:
Squashing for Centos images (both during testing and the actual build) has been disabled for quite some time, it was causing a number of problems:

1) it would frequently cause builds to hang (would hang the docker daemon)
2) there was inconsistent behavior introduced by squashing images in which file permissions or other filesystem attributes were behaving strangely.

I'm not in favor of re-enabling it, at least not in the infrastructure I maintain.

omron93:
@bparees Thanks for explaining this.

In PR testing I think squashing is not necessary. However, how openshift build and produce its images? They are not squashed?

bparees:
> In PR testing I think squashing is not necessary. However, how openshift build and produce its images? They are not squashed?

correct, the images we push to dockerhub are not squashed.  If we were going to squash the images we push to dockerhub, then it would definitely be necessary to do it as part of PR testing also, otherwise we are not testing what we publish and that would be bad (because, as i said, i've seen cases where squashing changes behavior).

omron93:
@bparees OK, so lets make PR testing consistent and I disable squashing also for non Openshift images.

Is someone using this functionality in Makefile? Fix it or remove it?

bparees:
The PR testing being run by us on our jenkins server has squashing disabled for all images, i believe.  Not sure what you guys may have setup on your side.

bparees:
@omron93 so squash behavior still happens by default if you just run the makefile, but our infrastructure explicitly disables the squashing via the SKIP_SQUASH=1 env variable.

at this point that's probably bad since it means people building this themselves will get squashed images, when we don't squash them ourselves.  so disabling it seems reasonable to me, but it's up to you guys at this point.. you're also welcome to pursue re-enabling it in your own build/push infrastructure.

hhorak:
I've separated squashing-by-default question to a separate issue:
https://github.com/sclorg/rhscl-container-ci/issues/5

Otherwise there seem to be no other issues, so LGTM.

PR #10, opened 2016-08-10
Allow to pull base image during build

omron93:
This PR adds option to update base image during build. So user can add "UPDATE_BASE=1" to `make` command and `docker build` will try to pull latest base image - see "--pull" option in https://docs.docker.com/engine/reference/commandline/build/

@hhorak @bparees Please test. I will create same PR for other repositories after this PR is complete and merged. Thanks.

### DIFF
--- a/hack/build.sh
+++ b/hack/build.sh
@@ -36,7 +36,10 @@ function docker_build_with_version {
   cp ${DOCKERFILE_PATH} "${DOCKERFILE_PATH}.version"
   git_version=$(git rev-parse --short HEAD)
   echo "LABEL io.openshift.builder-version=\"${git_version}\"" >> "${dockerfile}.version"
-  docker build -t ${IMAGE_NAME} -f "${dockerfile}.version" .
+  if [[ "${UPDATE_BASE}" == "1" ]]; then
+    BUILD_OPTIONS+=" --pull=true"
+  fi
+  docker build ${BUILD_OPTIONS} -t ${IMAGE_NAME} -f "${dockerfile}.version" .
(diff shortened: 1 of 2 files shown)


hhorak:
Sounds good to me. LGTM, it doesn't change default, so it should not cause any troubles.

bparees:
couple things:
1) you misaligned the indent on the makefile(nit)
2) you need to be careful when using this, because if you're trying to rebuild all the images including the base image, you do *not* want to specify this flag, you want to use the base image you just rebuilt locally, not the one that is in the registry and not updated.  (When we rebuild the base image, we also rebuild all the downstream images and retest them, to ensure the change to the base image has not broken any downstream image, before we push the new base image to dockerhub)

omron93:
> 1) you misaligned the indent on the makefile(nit)

Should be fixed. Thanks.

> ) you need to be careful when using this, because if you're trying to rebuild all the images including the base image, you do not want to specify this flag, you want to use the base image you just rebuilt locally, not the one that is in the registry and not updated. (When we rebuild the base image, we also rebuild all the downstream images and retest them, to ensure the change to the base image has not broken any downstream image, before we push the new base image to dockerhub)

This testing of downstream images is also applied during testing of PR for base image? (during testing of base image PR all dependent images are rebuilt)

bparees:
> This testing of downstream images is also applied during testing of PR for base image? (during testing of base image PR all dependent images are rebuilt)

if i recall correctly, yes, though it's been a while since i've looked at it.  it's handled within our jenkins job for base image PR testing.

PR #14, opened 2016-10-10
Pass image ID for testing

omron93:
When running `test/run` use image ID instead of IMAGE_NAME. This allows parallel testing of images (for example parallel testing of PRs in Jenkins).

@hhorak @bparees @praiskup Please take a look. I will prepare PRs for other repos later.

### DIFF
--- a/hack/build.sh
+++ b/hack/build.sh
@@ -13,6 +13,9 @@
 # OPENSHIFT_NAMESPACES - Which of available versions (subdirectories) should be
 #       put into openshift/ namespace.
 
+set -o errexit
+set -o pipefail
+
 OS=${1-$OS}
 VERSION=${2-$VERSION}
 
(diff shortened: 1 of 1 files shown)


hhorak:
This sounds like a good idea to me.

bparees:
lgtm

hack/build.sh
```diff
@@ -37,7 +37,10 @@ function docker_build_with_version {
   if [[ "${UPDATE_BASE}" == "1" ]]; then
     BUILD_OPTIONS+=" --pull=true"
   fi
-  docker build ${BUILD_OPTIONS} -t ${IMAGE_NAME} -f "${dockerfile}.version" .
+  local build_output=$(mktemp)
+  docker build ${BUILD_OPTIONS} -t ${IMAGE_NAME} -f "${dockerfile}.version" . | tee $build_output
```
praiskup: This probabyl requires `pipefail` set...  Otherwise we can miss `docker build` failure IMO.

praiskup:
Please consider documenting this in commit message, too.  Otherwise after fixing pipefail issue, it looks like nice improvement.  Thanks!

hack/build.sh
```diff
@@ -72,6 +75,7 @@ for dir in ${dirs}; do
   esac
 
   IMAGE_NAME="${NAMESPACE}${BASE_IMAGE_NAME}-${dir//./}-${OS}"
+  IMAGE_ID=""
```
omron93: OK, thanks.

hack/build.sh
omron93: > Otherwise we can miss docker build failure IMO.

@praiskup You are right. To catch failure, we have to fix `set -e` also, or not?

PR #16, opened 2016-11-01
remove force tag flag

bparees:


omron93:
@bparees Could you please proved some description?

I am afraid this error could appear:
```
Error response from daemon: Conflict: Tag centos/mariadb-100-centos7:latest is already set to image 5f0601d8685e, if you want to replace it, please use -f option
```

bparees:
The flag was removed in docker 1.12, you'll get an error if you specify it.

Ben Parees | OpenShift

PR #23, opened 2017-02-09
Port dockerfile for fedora and include it as addition

hhorak:
This only demonstrates that we can use the same sources for the container for SCL and Fedora packages. It does not influence any existing sources.


hhorak:
Since this change does not influence any existing sources, merging.

PR #25, opened 2017-02-13
Fix tests for docker 1.12 (changed docker exec parsing)

omron93:
Upgrade from docker 1.10 to docker 1.12 changed format of parsing `docker exec` arguments.

So `docker exec $(get_cid "$id") bash -c mysql <<< "SELECT 1;"` is waiting forever. Quotes or `-i` is required to make test passing.

EDIT: `assert_local_access` test was not working right with docker 1.10. (`mysql` got no input, so it succeded)

### DIFF
--- a/10.0/test/run
+++ b/10.0/test/run
@@ -204,7 +204,7 @@ function assert_login_access() {
 
 function assert_local_access() {
   local id="$1" ; shift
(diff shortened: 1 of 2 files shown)


omron93:
Tests are passing and this is really small change, so merging this PR to make build scripts working.

PR #26, opened 2017-02-16
Added extension point 

andreasschramm:
Hi, please consider adding extension points for users.

### DIFF
--- a/10.1/root/usr/bin/run-mysqld
+++ b/10.1/root/usr/bin/run-mysqld
@@ -6,6 +6,14 @@ set -eu
 
 [ -f ${CONTAINER_SCRIPTS_PATH}/validate-variables.sh ] && source ${CONTAINER_SCRIPTS_PATH}/validate-variables.sh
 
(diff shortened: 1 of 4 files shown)


hhorak:
Thanks for the contribution, we had several discussions about how to make this generally and it currently seems like using s2i approach might be better, see #34 for more details.

andreasschramm:
I was about to ask about that, i have seen your pull request, and agree it might be a better approach. 
I have yet to find out the best way to do some other things, yum and whatever.

hhorak:
Sorry, my previous comment mentioned #34, but I wanted to point to #32.. but you probably found the right one.. I've also added an example of a Dockerfile that changes USER and thus can install additional packages -- https://github.com/hhorak/mariadb-container/blob/s2i-support/10.1/test/Dockerfile

andreasschramm:
thanks, i was unsure on to how to do it "the right way", dockerfile or assemble. Still finding out the best  way to provide such functionality in a userfriendly way.

PR #31, opened 2017-04-12
add rhel6 10.1 build

tchughesiv:
some shell & python script changes required for compatibility w/ all 3 OS builds.

rhel6 only supports mariadb-10.1 so far.

### DIFF
--- a/10.1/Dockerfile.rhel6
+++ b/10.1/Dockerfile.rhel6
@@ -1,4 +1,4 @@
-FROM registry.access.redhat.com/rhel7
+FROM registry.access.redhat.com/rhel6
 
 # MariaDB image for OpenShift.
 #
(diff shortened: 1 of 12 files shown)


hhorak:
This looks good, but I miss the general idea why we need this :) I mean when we have mariadb 10.1 in the containers, what will be different when running the RHEL-6 and RHEL-7 based variants?

tchughesiv:
i think the main reason is a more seamless transition path (and comfort level) for customers/users.  For some, moving to containers is a big enough jump w/o also changing the OS.

Also, though maybe minor, there ARE differences (python & shell were 2 that i hit just creating rhel6 versions for scl) ... so the less you change in a migration, the less chance of unexpected things breaking.

I suspect these are some of the same reasons that we offer an official rhel6 image to begin with.

PR #32, opened 2017-04-19
WIP: Implement s2i support

hhorak:
This is very similar to the MongoDB PoC: https://github.com/sclorg/mongodb-container/pull/239/

It adds extending support using [source-to-image](https://github.com/openshift/source-to-image).

For example to build customized MariaDB database image `my-mariadb-centos7` with configuration in `~/image-configuration/` run:

```
$ s2i build ~/image-configuration/ centos/mariadb-101-centos7 my-mariadb-centos7
```

The directory passed to `s2i build` can contain these directories:
- `mysql-cfg/`
  - when starting the container, files from this directory will be used as a configuration for the `mysqld` daemon
    - `envsubst` command is run on this file to still allow customization of the image using environmental variables

- `mysql-pre-init/`
  - shell scripts (`*.sh`) available in this directory are sourced before `mysqld` daemon is started

- `mysql-init/`
  - shell scripts (`*.sh`) available in this directory are sourced when `mysqld` daemon is started locally
    - in this phase, use `${mysql_flags}` to connect to the locally running daemon, for example `mysql $mysql_flags < dump.sql`

Variables that can be used in the scripts provided to s2i:

- `$mysql_flags` -- arguments for the `mysql` tool that will connect to the locally running `mysqld` during initialization
- `$MYSQL_RUNNING_AS_MASTER` -- variable defined when the container is run with `run-mysqld-master` command
- `$MYSQL_RUNNING_AS_SLAVE` -- variable defined when the container is run with `run-mysqld-slave` command
- `$MYSQL_DATADIR_FIRST_INIT` -- variable defined when the container was initialized from the empty data dir

During `s2i build` all provided files are copied into `/opt/app-root/src` directory into the resulting image. If some configuration files are present in the destination directory, files with the same name are overwritten. Also only one file with the same name can be used for customization and user provided files are preferred over default files in `/usr/share/container-scripts/mysql/`- so it is possible to overwrite them.

Same configuration directory structure can be used to customize the image every time the image is started using `docker run`. The directory has to be mounted into `/opt/app-root/src/` in the image (`-v ./image-configuration/:/opt/app-root/src/`). This overwrites customization built into the image.


hhorak:
FYI @omron93  @praiskup @pkubatrh @bparees

pkubatrh:
I see some warnings when running the fix-permissions script on the s2i config files:

---> Installing application source ...
E0420 10:00:13.154889 15310 util.go:91] chgrp: changing group of './mysql-cfg/myconfig.cnf': Operation not permitted
E0420 10:00:13.154971 15310 util.go:91] chgrp: changing group of './mysql-cfg': Operation not permitted
E0420 10:00:13.154990 15310 util.go:91] chgrp: changing group of './mysql-data/init.sql': Operation not permitted
E0420 10:00:13.155050 15310 util.go:91] chgrp: changing group of './mysql-data': Operation not permitted
E0420 10:00:13.155129 15310 util.go:91] chgrp: changing group of './mysql-init/80-add-arbitrary-users.sh': Operation not permitted
E0420 10:00:13.155152 15310 util.go:91] chgrp: changing group of './mysql-init/90-init-db.sh': Operation not permitted
E0420 10:00:13.155173 15310 util.go:91] chgrp: changing group of './mysql-init': Operation not permitted
E0420 10:00:13.155190 15310 util.go:91] chgrp: changing group of './mysql-pre-init/80-check-arbitrary-users.sh': Operation not permitted
E0420 10:00:13.155208 15310 util.go:91] chgrp: changing group of './mysql-pre-init': Operation not permitted

omron93:
@hhorak Please take a look in updated `process_extending_files` function - https://github.com/omron93/mongodb-container/blob/da8700451761fa62caf5591292fad549060e4551/3.2/root/usr/share/container-scripts/mongodb/common.sh#L130

hhorak:
@tomastomecek suggested we should document extending images in the Dockerfile, so one can e.g. change USER to 0 to be able to install RPMs..

pkubatrh:
>suggested we should document extending images in the Dockerfile, so one can e.g. change USER to 0 to be able to install RPMs

I am not sure what is meant here by 'extending images in the Dockerfile' and how it is connected to s2i.
If it is a case of trying to do an s2i build that would install additional RPMs as an alternative to extending the image through FROM ... I do not think Openshift allows this.

bparees:
> I am not sure what is meant here by 'extending images in the Dockerfile' and how it is connected to s2i.
If it is a case of trying to do an s2i build that would install additional RPMs as an alternative to extending the image through FROM ... I do not think Openshift allows this.

it means create a docker strategy build in openshift that has a dockerfile which starts FROM this image, sets USER root, yum installs whatever you need, then sets USER xxxx (whatever the default user for the image was originally).

That will produce a customized version of this image, w/ the additional packages.

You can then use that image as your s2i builder image, which can you use in a normal s2i strategy build.

hhorak:
I think something like this should be enough:
```
FROM rhscl/mariadb-101-rhel7
COPY test-app /opt/app-root/src
```

hhorak:
@omron93 thanks, `<<<` is definitely better.

hhorak:
@tomastomecek There is now an example `Dockerfile` as well, that demonstrates installation of an additional RPM: https://github.com/hhorak/mariadb-container/blob/s2i-support/10.1/test/Dockerfile

10.1/test/Dockerfile
```diff
@@ -0,0 +1,11 @@
+# Extending the image using s2i concept can be done with a new Dockerfile as well
+# This way we can for example install additional RPM packages
+FROM __IMAGE_NAME__
+USER 0
+RUN INSTALL_PKGS="openssh-server" && \
+    yum install -y --setopt=tsflags=nodocs $INSTALL_PKGS && \
+    rpm -V $INSTALL_PKGS && \
+    yum clean all
+USER 27
+COPY test-app /opt/app-root/src
```
bparees: is this really necessary?  (having this file/"test"?)  it's docker image 101 (mentioning how to do this in the readme seems reasonable, providing a dockerfile to do it seems unnecessary, having a test that builds the dockerfile seems beyond unnecessary)

hhorak: Definitely not necessary. Although it happened to us recently that one image that was built on top of ours got broken because we changed something we believed was an implementation detail.. so yeah, it's hard to find the line..

TomasTomecek: I'm definitely with Honza here: having test for this ensures it's working and supported. It's also a very nice example.

hhorak:
This PR has been re-worked from scratch and continues in #45.

PR #34, opened 2017-04-25
Add logic for upgrading datadir

hhorak:
MySQL and MariaDB use versions that consist of three numbers X.Y.Z (e.g. 5.6.23).
For version changes in Z part, the server's binary data format stays compatible and thus no
special upgrade procedure is needed. For upgrades from X.Y to X.Y+1, consider doing manual
steps as described at
https://mariadb.com/kb/en/library/upgrading-from-mariadb-101-to-mariadb-102/

Skipping versions like from X.Y to X.Y+2 or downgrading to lower version is not supported;
the only exception is ugrading from MariaDB 5.5 to MariaDB 10.0.

**Important**: Upgrading to a new version is always risky and users are expected to make a full
back-up of all data before.

A safer solution to upgrade is to dump all data using `mysqldump` or `mysqldbexport` and then
load the data using `mysql` or `mysqldbimport` into an empty (freshly initialized) database.

Another way of proceeding with the upgrade is starting the new version of the `mysqld` daemon
and run `mysql_upgrade` right after the start. This so called in-place upgrade is generally
faster for large data directory, but only possible if upgrading from the very previous version,
so skipping versions is not supported.

This container detects whether the data needs to be upgraded using `mysql_upgrade` and
we can control it by setting `MYSQL_UPGRADE` variable, which can have one or more of the following values:

 * `warn` -- If the data version can be determined and the data come from a different version
   of the daemon, a warning is printed but the container starts. This is the default value.
   Since historically the version file `mysql_upgrade_info` was not created, when using this option,
   the version file is created if not exist, but no `mysql_upgrade` will be called.
   However, this automatic creation will be removed after few months, since the version should be
   created on most deployments at that point.
 * `auto` -- `mysql_upgrade` is run at the beginning of the container start, if the data version
   can be determined and the data come with the very previous version. A warning is printed if the data
   come from even older or newer version. This value effectively enables automatic upgrades,
   but it is always risky and users should still back-up all the data before starting the newer container.
   Set this option only if you have very good back-ups at any moment and you are fine to fail-over
   from the back-up.
 * `force` -- `mysql_upgrade` is run right after the daemon has started, no matter what version the data
   come from. This is also the way to create the missing version file `mysql_upgrade_info` if not present
   in the root of the data directory; this file holds information about the version of the data.
 * `optimize` -- runs `mysqlcheck --optimize` before starting the mysqld daemon, no matter what version
   of the data is detected. It optimizes all the tables.
 * `analyze` -- runs `mysqlcheck --analyze` before starting the mysqld daemon, no matter what version
   of the data is detected. It analyzes all the tables.
 * `disable` -- nothing is done regarding data directory version.

Multiple values are separated by comma and run in-order, e.g. `MYSQL_UPGRADE="optimize,analyze"`.

### DIFF
--- a/10.0/root/usr/share/container-scripts/mysql/README.md
+++ b/10.0/root/usr/share/container-scripts/mysql/README.md
@@ -292,19 +292,21 @@ we can control it by setting `MYSQL_UPGRADE` variable, which can have one or mor
    the version file is created if not exist, but no `mysql_upgrade` will be called.
    However, this automatic creation will be removed after few months, since the version should be
    created on most deployments at that point.
- * `auto` -- `mysql_upgrade` is run at the beginning of the container start, if the data version
-   can be determined and the data come with the very previous version. A warning is printed if the data
-   come from even older or newer version. This value effectively enables automatic upgrades,
+ * `auto` -- `mysql_upgrade` is run at the beginning of the container start, when the local
+   daemon is running, but only if the data version can be determined and the data come
+   with the very previous version. A warning is printed if the data come from even older
+   or newer version. This value effectively enables automatic upgrades,
    but it is always risky and users should still back-up all the data before starting the newer container.
    Set this option only if you have very good back-ups at any moment and you are fine to fail-over
    from the back-up.
- * `force` -- `mysql_upgrade` is run right after the daemon has started, no matter what version the data
-   come from. This is also the way to create the missing version file `mysql_upgrade_info` if not present
+ * `force` -- `mysql_upgrade --force` is run at the beginning of the container start, when the local
+   daemon is running, no matter what version of the daemon the data come from.
+   This is also the way to create the missing version file `mysql_upgrade_info` if not present
    in the root of the data directory; this file holds information about the version of the data.
- * `optimize` -- runs `mysqlcheck --optimize` before starting the mysqld daemon, no matter what version
-   of the data is detected. It optimizes all the tables.
- * `analyze` -- runs `mysqlcheck --analyze` before starting the mysqld daemon, no matter what version
-   of the data is detected. It analyzes all the tables.
+ * `optimize` -- runs `mysqlcheck --optimize` at the beginning of the container start, when the local
+   daemon is running, no matter what version of the data is detected. It optimizes all the tables.
+ * `analyze` -- runs `mysqlcheck --analyze` at the beginning of the container start, when the local
+   daemon is running, no matter what version of the data is detected. It analyzes all the tables.
  * `disable` -- nothing is done regarding data directory version.
 
 Multiple values are separated by comma and run in-order, e.g. `MYSQL_UPGRADE="optimize,analyze"`.
--- a/10.0/root/usr/share/container-scripts/mysql/README.md
+++ b/10.0/root/usr/share/container-scripts/mysql/README.md
@@ -284,32 +284,34 @@ faster for large data directory, but only possible if upgrading from the very pr
 so skipping versions is not supported.
 
 This container detects whether the data needs to be upgraded using `mysql_upgrade` and
-we can control it by setting `MYSQL_UPGRADE` variable, which can have one or more of the following values:
+we can control it by setting `MYSQL_DATADIR_ACTION` variable, which can have one or more of the following values:
 
- * `warn` -- If the data version can be determined and the data come from a different version
+ * `upgrade-warn` -- If the data version can be determined and the data come from a different version
    of the daemon, a warning is printed but the container starts. This is the default value.
    Since historically the version file `mysql_upgrade_info` was not created, when using this option,
    the version file is created if not exist, but no `mysql_upgrade` will be called.
    However, this automatic creation will be removed after few months, since the version should be
    created on most deployments at that point.
- * `auto` -- `mysql_upgrade` is run at the beginning of the container start, when the local
+ * `upgrade-auto` -- `mysql_upgrade` is run at the beginning of the container start, when the local
    daemon is running, but only if the data version can be determined and the data come
    with the very previous version. A warning is printed if the data come from even older
    or newer version. This value effectively enables automatic upgrades,
    but it is always risky and users should still back-up all the data before starting the newer container.
    Set this option only if you have very good back-ups at any moment and you are fine to fail-over
    from the back-up.
- * `force` -- `mysql_upgrade --force` is run at the beginning of the container start, when the local
+ * `upgrade-force` -- `mysql_upgrade --force` is run at the beginning of the container start, when the local
    daemon is running, no matter what version of the daemon the data come from.
    This is also the way to create the missing version file `mysql_upgrade_info` if not present
    in the root of the data directory; this file holds information about the version of the data.
- * `optimize` -- runs `mysqlcheck --optimize` at the beginning of the container start, when the local
-   daemon is running, no matter what version of the data is detected. It optimizes all the tables.
- * `analyze` -- runs `mysqlcheck --analyze` at the beginning of the container start, when the local
-   daemon is running, no matter what version of the data is detected. It analyzes all the tables.
+
+There are also some other actions that you may want to run at the beginning of the container start,
+when the local daemon is running, no matter what version of the data is detected:
+
+ * `optimize` -- runs `mysqlcheck --optimize`. It optimizes all the tables.
+ * `analyze` -- runs `mysqlcheck --analyze`. It analyzes all the tables.
  * `disable` -- nothing is done regarding data directory version.
 
-Multiple values are separated by comma and run in-order, e.g. `MYSQL_UPGRADE="optimize,analyze"`.
+Multiple values are separated by comma and run in-order, e.g. `MYSQL_DATADIR_ACTION="optimize,analyze"`.
 
 
 Changing the replication binlog_format
--- a/10.0/root/usr/share/container-scripts/mysql/README.md
+++ b/10.0/root/usr/share/container-scripts/mysql/README.md
@@ -260,6 +260,56 @@ Such a directory `sslapp` can then be mounted into the container with -v,
 or a new container image can be built using s2i.
 
 
+Upgrading and data directory version checking
+---------------------------------------------
+
+MySQL and MariaDB use versions that consist of three numbers X.Y.Z (e.g. 5.6.23).
+For version changes in Z part, the server's binary data format stays compatible and thus no
+special upgrade procedure is needed. For upgrades from X.Y to X.Y+1, consider doing manual
+steps as described at
+https://mariadb.com/kb/en/library/upgrading-from-mariadb-55-to-mariadb-100/
+
+Skipping versions like from X.Y to X.Y+2 or downgrading to lower version is not supported;
+the only exception is ugrading from MariaDB 5.5 to MariaDB 10.0.
+
+**Important**: Upgrading to a new version is always risky and users are expected to make a full
+back-up of all data before.
+
+A safer solution to upgrade is to dump all data using `mysqldump` or `mysqldbexport` and then
+load the data using `mysql` or `mysqldbimport` into an empty (freshly initialized) database.
+
+Another way of proceeding with the upgrade is starting the new version of the `mysqld` daemon
+and run `mysql_upgrade` right after the start. This so called in-place upgrade is generally
+faster for large data directory, but only possible if upgrading from the very previous version,
+so skipping versions is not supported.
+
+This container detects whether the data needs to be upgraded using `mysql_upgrade` and
+we can control it by setting `MYSQL_UPGRADE` variable, which can have one or more of the following values:
+
+ * `warn` -- If the data version can be determined and the data come from a different version
+   of the daemon, a warning is printed but the container starts. This is the default value.
+   Since historically the version file `mysql_upgrade_info` was not created, when using this option,
+   the version file is created if not exist, but no `mysql_upgrade` will be called.
+   However, this automatic creation will be removed after few months, since the version should be
+   created on most deployments at that point.
+ * `auto` -- `mysql_upgrade` is run at the beginning of the container start, if the data version
+   can be determined and the data come with the very previous version. A warning is printed if the data
+   come from even older or newer version. This value effectively enables automatic upgrades,
+   but it is always risky and users should still back-up all the data before starting the newer container.
+   Set this option only if you have very good back-ups at any moment and you are fine to fail-over
+   from the back-up.
+ * `force` -- `mysql_upgrade` is run right after the daemon has started, no matter what version the data
+   come from. This is also the way to create the missing version file `mysql_upgrade_info` if not present
+   in the root of the data directory; this file holds information about the version of the data.
+ * `optimize` -- runs `mysqlcheck --optimize` before starting the mysqld daemon, no matter what version
+   of the data is detected. It optimizes all the tables.
+ * `analyze` -- runs `mysqlcheck --analyze` before starting the mysqld daemon, no matter what version
+   of the data is detected. It analyzes all the tables.
+ * `disable` -- nothing is done regarding data directory version.
+
+Multiple values are separated by comma and run in-order, e.g. `MYSQL_UPGRADE="optimize,analyze"`.
+
+
 Changing the replication binlog_format
 --------------------------------------
 Some applications may wish to use `row` binlog_formats (for example, those built
--- a/10.1/root/usr/share/container-scripts/mysql/README.md
+++ b/10.1/root/usr/share/container-scripts/mysql/README.md
@@ -292,19 +292,21 @@ we can control it by setting `MYSQL_UPGRADE` variable, which can have one or mor
    the version file is created if not exist, but no `mysql_upgrade` will be called.
    However, this automatic creation will be removed after few months, since the version should be
    created on most deployments at that point.
- * `auto` -- `mysql_upgrade` is run at the beginning of the container start, if the data version
-   can be determined and the data come with the very previous version. A warning is printed if the data
-   come from even older or newer version. This value effectively enables automatic upgrades,
+ * `auto` -- `mysql_upgrade` is run at the beginning of the container start, when the local
+   daemon is running, but only if the data version can be determined and the data come
+   with the very previous version. A warning is printed if the data come from even older
+   or newer version. This value effectively enables automatic upgrades,
    but it is always risky and users should still back-up all the data before starting the newer container.
    Set this option only if you have very good back-ups at any moment and you are fine to fail-over
    from the back-up.
- * `force` -- `mysql_upgrade` is run right after the daemon has started, no matter what version the data
-   come from. This is also the way to create the missing version file `mysql_upgrade_info` if not present
+ * `force` -- `mysql_upgrade --force` is run at the beginning of the container start, when the local
+   daemon is running, no matter what version of the daemon the data come from.
+   This is also the way to create the missing version file `mysql_upgrade_info` if not present
    in the root of the data directory; this file holds information about the version of the data.
- * `optimize` -- runs `mysqlcheck --optimize` before starting the mysqld daemon, no matter what version
-   of the data is detected. It optimizes all the tables.
- * `analyze` -- runs `mysqlcheck --analyze` before starting the mysqld daemon, no matter what version
-   of the data is detected. It analyzes all the tables.
+ * `optimize` -- runs `mysqlcheck --optimize` at the beginning of the container start, when the local
+   daemon is running, no matter what version of the data is detected. It optimizes all the tables.
+ * `analyze` -- runs `mysqlcheck --analyze` at the beginning of the container start, when the local
+   daemon is running, no matter what version of the data is detected. It analyzes all the tables.
  * `disable` -- nothing is done regarding data directory version.
 
 Multiple values are separated by comma and run in-order, e.g. `MYSQL_UPGRADE="optimize,analyze"`.
--- a/10.1/root/usr/share/container-scripts/mysql/README.md
+++ b/10.1/root/usr/share/container-scripts/mysql/README.md
@@ -284,32 +284,34 @@ faster for large data directory, but only possible if upgrading from the very pr
 so skipping versions is not supported.
 
 This container detects whether the data needs to be upgraded using `mysql_upgrade` and
-we can control it by setting `MYSQL_UPGRADE` variable, which can have one or more of the following values:
+we can control it by setting `MYSQL_DATADIR_ACTION` variable, which can have one or more of the following values:
 
- * `warn` -- If the data version can be determined and the data come from a different version
+ * `upgrade-warn` -- If the data version can be determined and the data come from a different version
    of the daemon, a warning is printed but the container starts. This is the default value.
    Since historically the version file `mysql_upgrade_info` was not created, when using this option,
    the version file is created if not exist, but no `mysql_upgrade` will be called.
    However, this automatic creation will be removed after few months, since the version should be
    created on most deployments at that point.
- * `auto` -- `mysql_upgrade` is run at the beginning of the container start, when the local
+ * `upgrade-auto` -- `mysql_upgrade` is run at the beginning of the container start, when the local
    daemon is running, but only if the data version can be determined and the data come
    with the very previous version. A warning is printed if the data come from even older
    or newer version. This value effectively enables automatic upgrades,
    but it is always risky and users should still back-up all the data before starting the newer container.
    Set this option only if you have very good back-ups at any moment and you are fine to fail-over
    from the back-up.
- * `force` -- `mysql_upgrade --force` is run at the beginning of the container start, when the local
+ * `upgrade-force` -- `mysql_upgrade --force` is run at the beginning of the container start, when the local
    daemon is running, no matter what version of the daemon the data come from.
    This is also the way to create the missing version file `mysql_upgrade_info` if not present
    in the root of the data directory; this file holds information about the version of the data.
- * `optimize` -- runs `mysqlcheck --optimize` at the beginning of the container start, when the local
-   daemon is running, no matter what version of the data is detected. It optimizes all the tables.
- * `analyze` -- runs `mysqlcheck --analyze` at the beginning of the container start, when the local
-   daemon is running, no matter what version of the data is detected. It analyzes all the tables.
+
+There are also some other actions that you may want to run at the beginning of the container start,
+when the local daemon is running, no matter what version of the data is detected:
+
+ * `optimize` -- runs `mysqlcheck --optimize`. It optimizes all the tables.
+ * `analyze` -- runs `mysqlcheck --analyze`. It analyzes all the tables.
  * `disable` -- nothing is done regarding data directory version.
 
-Multiple values are separated by comma and run in-order, e.g. `MYSQL_UPGRADE="optimize,analyze"`.
+Multiple values are separated by comma and run in-order, e.g. `MYSQL_DATADIR_ACTION="optimize,analyze"`.
 
 
 Changing the replication binlog_format
--- a/10.1/root/usr/share/container-scripts/mysql/README.md
+++ b/10.1/root/usr/share/container-scripts/mysql/README.md
@@ -260,6 +260,56 @@ Such a directory `sslapp` can then be mounted into the container with -v,
 or a new container image can be built using s2i.
 
 
+Upgrading and data directory version checking
+---------------------------------------------
+
+MySQL and MariaDB use versions that consist of three numbers X.Y.Z (e.g. 5.6.23).
+For version changes in Z part, the server's binary data format stays compatible and thus no
+special upgrade procedure is needed. For upgrades from X.Y to X.Y+1, consider doing manual
+steps as described at
+https://mariadb.com/kb/en/library/upgrading-from-mariadb-100-to-mariadb-101/
+
+Skipping versions like from X.Y to X.Y+2 or downgrading to lower version is not supported;
+the only exception is ugrading from MariaDB 5.5 to MariaDB 10.0.
+
+**Important**: Upgrading to a new version is always risky and users are expected to make a full
+back-up of all data before.
+
+A safer solution to upgrade is to dump all data using `mysqldump` or `mysqldbexport` and then
+load the data using `mysql` or `mysqldbimport` into an empty (freshly initialized) database.
+
+Another way of proceeding with the upgrade is starting the new version of the `mysqld` daemon
+and run `mysql_upgrade` right after the start. This so called in-place upgrade is generally
+faster for large data directory, but only possible if upgrading from the very previous version,
+so skipping versions is not supported.
+
+This container detects whether the data needs to be upgraded using `mysql_upgrade` and
+we can control it by setting `MYSQL_UPGRADE` variable, which can have one or more of the following values:
+
+ * `warn` -- If the data version can be determined and the data come from a different version
+   of the daemon, a warning is printed but the container starts. This is the default value.
+   Since historically the version file `mysql_upgrade_info` was not created, when using this option,
+   the version file is created if not exist, but no `mysql_upgrade` will be called.
+   However, this automatic creation will be removed after few months, since the version should be
+   created on most deployments at that point.
+ * `auto` -- `mysql_upgrade` is run at the beginning of the container start, if the data version
+   can be determined and the data come with the very previous version. A warning is printed if the data
+   come from even older or newer version. This value effectively enables automatic upgrades,
+   but it is always risky and users should still back-up all the data before starting the newer container.
+   Set this option only if you have very good back-ups at any moment and you are fine to fail-over
+   from the back-up.
+ * `force` -- `mysql_upgrade` is run right after the daemon has started, no matter what version the data
+   come from. This is also the way to create the missing version file `mysql_upgrade_info` if not present
+   in the root of the data directory; this file holds information about the version of the data.
+ * `optimize` -- runs `mysqlcheck --optimize` before starting the mysqld daemon, no matter what version
+   of the data is detected. It optimizes all the tables.
+ * `analyze` -- runs `mysqlcheck --analyze` before starting the mysqld daemon, no matter what version
+   of the data is detected. It analyzes all the tables.
+ * `disable` -- nothing is done regarding data directory version.
+
+Multiple values are separated by comma and run in-order, e.g. `MYSQL_UPGRADE="optimize,analyze"`.
+
+
 Changing the replication binlog_format
 --------------------------------------
 Some applications may wish to use `row` binlog_formats (for example, those built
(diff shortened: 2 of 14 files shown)


hhorak:
@bparees @praiskup @pkubatrh FYI
This is a first draft to make the container a bit more bullet-proof. On the other hand, the 'auto' mode might be a bit risky, since it's something we don't do automatically in RPM cases (in comparison to Ubuntu, who do run `mysql_upgrade` automatically), so any feedback welcome..

10.1/root/usr/bin/run-mysqld-slave
```diff
@@ -48,8 +48,14 @@ mysql $mysql_flags <<EOSQL
...
-log_info 'Sourcing post-init.sh ...'
-[ -f ${CONTAINER_SCRIPTS_PATH}/post-init.sh ] && source ${CONTAINER_SCRIPTS_PATH}/post-init.sh
+if [ -f ${CONTAINER_SCRIPTS_PATH}/post-init.sh ]; then
+  log_info 'Sourcing post-init.sh ...'
+  source ${CONTAINER_SCRIPTS_PATH}/post-init.sh
+fi
+if [ -f ${CONTAINER_SCRIPTS_PATH}/check-upgrade.sh ]; then
+  log_info 'Sourcing check-upgrade.sh ...'
+  source ${CONTAINER_SCRIPTS_PATH}/check-upgrade.sh
+fi
```
praiskup: It seems like we should "upgrade" first and then do "post-init.sh".

10.1/root/usr/share/container-scripts/mysql/common.sh
```diff
@@ -92,6 +94,7 @@ function initialize_database() {
   # Using empty --basedir to work-around https://bugzilla.redhat.com/show_bug.cgi?id=1406391
   mysql_install_db --rpm --datadir=$MYSQL_DATADIR --basedir=''
   start_local_mysql "$@"
+  mysql_upgrade --socket=/tmp/mysql.sock
```
praiskup: As I'm not familiar with mysql/mariadb "upgrading" procedures, I'm curious why this method changed...  Can we have small comment above this command?

10.1/root/usr/share/container-scripts/mysql/common.sh
```diff
@@ -162,3 +165,63 @@ function wait_for_mysql_master() {
     sleep 1
   done
 }
+
+# Converts string version to the integer format (5.5 is converted to 505,
+# 10.1 is converted into 1001, etc.
+function version2number() {
+  local version=0
```
praiskup: I we have bash (we have I believe), this would be simpler:
```bash
printf %d%02d ${1%%.*} ${1##*.}
```

10.1/root/usr/share/container-scripts/mysql/common.sh
```diff
@@ -162,3 +165,63 @@ function wait_for_mysql_master() {
...
+  local version_major=$(echo "$version_text" | sed -e 's/\([0-9]*\)\.\([0-9]*\)\..*$/\1/')
+  local version_minor=$(echo "$version_text" | sed -e 's/\([0-9]*\)\.\([0-9]*\)\..*$/\2/')
+  if [[ $version_major =~ ^[0-9]+$ ]] && [[ $version_minor =~ ^[0-9]+$ ]] ; then
+    version=$((version_major*100+version_minor))
+  fi
+  if [ "$version" -eq 0 ] ; then
+    echo "Error: version '$version_text' cannot be converted to the number."
+    echo "0"
+  fi
+  echo "${version}"
```
praiskup: This is personal preference, but I can't resist -> I really like this pattern:
```bash
versionNumber() {
# Do something...
# and then set global variable $<function_name>_result (versionNumber_result)
printf -v versionNumber_result %d%02d ${1%%.*} ${1##*.}
}
```
By this pattern, you avoid additional shell forks and error prone statements like:
```bash
variable=`method with params`
```

10.1/root/usr/share/container-scripts/mysql/common.sh
```diff
@@ -162,3 +165,63 @@ function wait_for_mysql_master() {
...
+}
+
+# Prints version of the mysqld that is currently available (string)
+function mysqld_version() {
+  ${MYSQL_PREFIX}/libexec/mysqld -V | awk '{print $3}'
```
praiskup: On Fedora I get: 10.1.21-MariaDB

10.1/root/usr/share/container-scripts/mysql/common.sh
praiskup: Ah, we likely do more than what is in function description ... :)

10.1/root/usr/share/container-scripts/mysql/common.sh
```diff
@@ -162,3 +165,63 @@ function wait_for_mysql_master() {
...
+  echo "$datadir/mysql_upgrade_info"
+}
+
+# Gets version of the current daemon and stores it into the appropriate file in the datadir
+function store_mysqld_version() {
+  local datadir="$1"
+  local version=$(mysqld_version)
+  local upgrade_info_file=$(get_mysql_upgrade_info_file "$datadir")
+  echo "${version}" > "${upgrade_info_file}"
+  log_info "Storing version '${version}' information into the data dir '${upgrade_info_file}'"
```
praiskup: Please reverse the two commands above, or do `s/Storing/Stored/`.

praiskup: Ah, this function seems to be unused...

10.1/root/usr/share/container-scripts/mysql/check-upgrade.sh
```diff
@@ -0,0 +1,47 @@
+check_datadir_version() {
+  local datadir="$1"
+  local datadir_version=$(get_datadir_version "$datadir")
+  local mysqld_version=$(mysqld_compat_version)
+
+  case "${MYSQL_UPGRADE}" in
```
praiskup: Double quotes are not needed here.

10.1/root/usr/share/container-scripts/mysql/README.md
```diff
@@ -124,6 +124,42 @@ variable with the full path of the file you wish to use. For example, the defaul
 location is `/etc/my.cnf` but you can change it to `/etc/mysql/my.cnf` by setting
  `MYSQL_DEFAULTS_FILE=/etc/mysql/my.cnf`
 
+Upgrading and datadir checking
+------------------------------
+MySQL and MariaDB use versions that consist of three numbers (e.g. 5.6.23). When two
+versions are different only in the last part, they are compatible, but when upgrading
```
praiskup: Ah, I read this like dozen of times, ... :) and now I get it right...   I understood that statement like 'when two _numbers_ are different only in the last part', so I would suggest:
" ... that consists of three numbers `X.Y.Z` (e.g. 5.6.23).  For version changes in `Z` part, the server's binary data format stays compatible and thus no special upgrade procedure is needed.  For upgrades from `X.Y` to `X.Y+1`,  we need to consider upgrading ... "

  co

praiskup: I would document what happens when `X` is bumped, ...

10.1/root/usr/share/container-scripts/mysql/README.md
```diff
@@ -124,6 +124,42 @@ variable with the full path of the file you wish to use. For example, the defaul
...
+from e.g. 5.6 to 5.7 or from 10.0 to 10.1, we need to consider upgrading as described
+in https://dev.mysql.com/doc/refman/5.7/en/upgrading-from-previous-series.html.
+
```
praiskup: `s/consider upgrading as described/consider doing manual steps as described/` ?

10.1/root/usr/share/container-scripts/mysql/README.md
```diff
@@ -124,6 +124,42 @@ variable with the full path of the file you wish to use. For example, the defaul
...
+Another way of proceeding with the upgrade is starting the new version of the `mysqld` daemon
+and run `mysql_upgrade` right after the start. This so called in-place upgrade is only possible
+if upgrading from the very previous version, so skipping versions is not possible.
+
+This container tries to detect whether the data needs to be upgraded using `mysql_upgrade` and
+we can control it by setting `MYSQL_UPGRADE` variable, which can have the following values:
+
+ * `warn` -- If the data version can be determined and the data come from a different version
+   of the daemon, a warning is printed but the container starts. This is the default value.
```
praiskup: Is it really safe to start `mysqld` against incompatible data version _by default_?

10.1/root/usr/share/container-scripts/mysql/README.md
```diff
@@ -124,6 +124,42 @@ variable with the full path of the file you wish to use. For example, the defaul
...
+ * `auto` -- `mysql_upgrade` is run at the beginning of the container start, if the data version
+   can be determined and the data come with the very previous version. A warning is printed if the data
+   come from even older or newer version. This value effectively enables automatic upgrades,
+   but it is always risky and users should still back-up all the data before starting the newer container.
+   Set this option only if you have very good back-ups at any moment and you are fine to fail-over
+   from the back-up.
+ * `force` -- `mysql_upgrade` is run right after the daemon has started, no matter what version the data
+   come from. This can be also used to creating the `mysql_upgrade_info` file if not present in the root
```
praiskup: s/used to creating/used to create/

But I wouldn't probably mention that this is "the way to create the missing version file" .. Perhaps this deserves another compat option ...?

10.1/root/usr/share/container-scripts/mysql/README.md
```diff
@@ -124,6 +124,42 @@ variable with the full path of the file you wish to use. For example, the defaul
...
+   of the data directory; this file holds information about the version of the data.
+ * `optimize` -- runs `mysqlcheck --optimize` before starting the mysqld daemon, no matter what version
+   of the data is detected. It optimizes all the tables.
```
praiskup: It would be nice if this was not exclusive option, so users could use:
`MYSQL_UPGRADE=optimize,force,analyze`
(if that makes sense)

10.1/root/usr/share/container-scripts/mysql/check-upgrade.sh
```diff
@@ -0,0 +1,47 @@
...
+          mysql_upgrade --socket=/tmp/mysql.sock
+        fi
+      else
+        log_info "MySQL server is version '${mysqld_version}' and datadir is version '${datadir_version}', which are incompatible."
+      fi
+      ;;
+
+    force)
+      log_info 'Running mysql_upgrade --force ...'
+      mysql_upgrade --socket=/tmp/mysql.sock --force
```
praiskup: Can we document this is safe ("no-op") if the version isn't changed?  OTOH, I'm afraid of situations where users will use `MYSQL_UPGRADE=force` temporarily (because of missing version file), and they forgot to remove this variable from configuration .... :/  then simple changing of image to "too new" version might end up with disaster...

10.1/root/usr/share/container-scripts/mysql/check-upgrade.sh
```diff
@@ -0,0 +1,47 @@
...
+      fi
+
+      if [ $(( ${datadir_version} + 1 )) -eq "${mysqld_version}" -o "${datadir_version}" -eq 505 -a "${mysqld_version}" -eq 1000 ] ; then
+        log_info "MySQL server is version '${mysqld_version}' and datadir is version '${datadir_version}'."
+        if [ "${MYSQL_UPGRADE}" == 'auto' ] ; then
+          log_info 'Running mysql_upgrade ...'
+          mysql_upgrade --socket=/tmp/mysql.sock
+        fi
+      else
+        log_info "MySQL server is version '${mysqld_version}' and datadir is version '${datadir_version}', which are incompatible."
```
praiskup: Why isn't failure expected here?

praiskup: Also, maybe we should info user about the possibilities (what is the name of the container that supports the upgrade from this datadir version, and that multiple upgrade steps might be needed e.g.).

praiskup:
Thanks for the progress!  I have some notes:
- what's the storage implication of "in place" upgrade, do we need some additional data or is that really "in situ"?
- could we (perhaps, if storage allows us) do the "in container" backup and restore it in case of failure?  Or would that be over-engineering?
- what is the expected time constraints (the downtime caused by "in-place upgrade")

While I really dislike the `N` -> `N+1` -> `N+2` upgrade approach (if some major version needs to be skipped -> which is the most likely scenario btw), it looks like in case of mariadb this "in-place" procedure has no real drawbacks to not have it implemented.

10.1/root/usr/bin/run-mysqld-slave
pkubatrh: Agreed. The user could run some script as part of post-init that could fail when the data is not upgraded.

10.1/root/usr/share/container-scripts/mysql/README.md
pkubatrh: Not that familiar with mariadb/mysql but the analyze and optimize options seem like something that would be useful even outside data upgrade.

10.1/root/usr/bin/run-mysqld-slave
hhorak: Sure, makes sense, I'll fix this.

10.1/root/usr/share/container-scripts/mysql/common.sh
hhorak: That's something I found during investigation, that `mysql_upgrade` also marks data properly with a version, so we can either add it manually or let `mysql_upgrade` do it, while it also checks the data dir, which in empty (freshly initialize) case is very quick and might find potential issues (which is not expected, but one never knows).

10.1/root/usr/share/container-scripts/mysql/common.sh
hhorak: Correct. mariadb uses version in format e.g. 10.0.12-MariaDB, so we need to remove the -MariaDB or any similar suffix as well. It might still be done better in bash though, so will try to fix..

10.1/root/usr/share/container-scripts/mysql/common.sh
hhorak: right, make sense to me

10.1/root/usr/share/container-scripts/mysql/common.sh
hhorak: It's replaced by mysql_upgrade as described above, so I'll probably remove this function at all..

10.1/root/usr/share/container-scripts/mysql/check-upgrade.sh
pkubatrh: If I understand it correctly all the `--force` option does is ignore the existence of a version file. I expect `mysql_upgrade` should still fail if it encounters some incompatibilities with the existing version.

10.1/root/usr/share/container-scripts/mysql/common.sh
praiskup: Makes sense, but comment would be nice.

10.1/root/usr/share/container-scripts/mysql/check-upgrade.sh
praiskup: But in case of "auto|warn", we do manual version checking (instead of delegating the checks  to `mysql_upgrade`).  At least we don't always run the `mysql_upgrade` -> so it seemed to me that it is somehow unsafe action.

10.1/root/usr/share/container-scripts/mysql/README.md
pkubatrh: I definitely would not do this by default since the warning will be missed (unless something breaks).
I have looked into how the image updates are managed in Openshift (via imagestreams) and it seems like as long as the user is using mariadb:latest in the imagestream and a new mariadb image (that can be of a different version) gets tagged by :latest, Openshift will just take down the old deployment and deploy the new version automatically.

hhorak:
> what's the storage implication of "in place" upgrade, do we need some additional data or is that really "in situ"?

If you mean space-wise, whether we need more space, then we should not need much more, to my knowledge; the upgrade is basically about changing some internal structures mostly, analyzing tables, etc.

hhorak:
> could we (perhaps, if storage allows us) do the "in container" backup and restore it in case of failure? Or would that be over-engineering?

That would bring new requirements on space in the volume and users would then either have another back-up somewhere else or would need to figure out how the back-up can be used. I'd say that conscious user backs-up data regularly and especially before upgrade, so going back and using back-up in case of troubles is something what users should be counted with always. I don't think we need to do this job for them.

hhorak:
> what is the expected time constraints (the downtime caused by "in-place upgrade")

Depending on the size, I expect up to minutes in case of reasonably big tables, but I don't have any specific data. My guess is that the data are primarily read and only some quick operation is proceeded on them, so maybe thinking about what is the speed of sequential read might give an idea..

hhorak:
> application container can wait ... I mean, with this approach the app deployment plan isn't changed and I'm curious whether it hasn't some side effects

The upgrade should IMO happen at the time the daemon is not listening outside, which means the case when `MYSQL_UPGRADE` is set to 'auto'. Then the application should be smart enough to wait for the database to come up. However, even if the database upgrade is run on live DB (listening outside), the data are locked for reading, so application will be able to know..

10.1/root/usr/share/container-scripts/mysql/check-upgrade.sh
hhorak: Sounds fine to me, thanks for the suggestions.

10.1/root/usr/share/container-scripts/mysql/check-upgrade.sh
hhorak: I should take a closer look into the code, but quick look at the behavior and into the man page it looks like it is effectively no-op.

10.1/root/usr/share/container-scripts/mysql/README.md
hhorak: Right, that makes sense to me as well.

10.1/root/usr/share/container-scripts/mysql/README.md
hhorak: I see your point, but on the other hand less options means simpler image.. so I'm not sure whether it is worth it..

10.1/root/usr/share/container-scripts/mysql/README.md
hhorak: There are users who know what they are doing and in that case starting a container and running `mysql_upgrade` right after it, even manually, is totally valid use case. But I see the point in that it maybe should not be the default.. On the other hand, RPMs currently do the same by default -- the newer daemon starts, but an error/warning is printed into the log and we don't have any reports for this. Anyway, the situation with the image stream makes it a bit complicated and less visible, so we might pick a different strategy here. I'm still not decided though.

10.1/root/usr/share/container-scripts/mysql/README.md
hhorak: ok, thx

10.1/root/usr/share/container-scripts/mysql/README.md
hhorak: Nice wording, I admit that the initial one was not understandable for me either.

10.1/root/usr/share/container-scripts/mysql/common.sh
hhorak: definitely, will add

10.1/root/usr/share/container-scripts/mysql/README.md
hhorak: Still thinking about this.. `mysql_upgrade`'s man page says this: "...also saves the MariaDB version number in a file named mysql_upgrade_info in the data directory. This is used to quickly check whether all tables have been checked for this release so that table-checking can be skipped..."

So the file presence means the data were checked for a particular version, it's not just information about a version.. and thus it makes sense to create it only after `mysql_upgrade` run. So maybe we should not include the recommendation of creating the file manually and rather advise to use `--force` for creating this file..

10.1/root/usr/share/container-scripts/mysql/check-upgrade.sh
hhorak: I've checked the code and it really is a no-op.

hhorak: Taking back, if we use '--force', than it is not a no-op.. it will call `mysqlcheck` which runs optimize/check/repair tables..

hhorak: But my expectation is that even running it more times should be safe..

10.1/root/usr/share/container-scripts/mysql/README.md
pkubatrh: In my opinion running `mysql_upgrade` by default might actually be a bit better (=less evil) in this case, since at least this way the fail would occur at the moment of upgrade and not far off in the future as with the ignored warning (and probably be less/not-at-all destructive?).
Best case would be if there were no :latest tag in the imagestream for DB images but I am not sure if that is possible or even wanted by Openshift.

10.1/root/usr/share/container-scripts/mysql/README.md
praiskup: Does that mean that if the file is not possible to be created (some table is not checked against particular mariadb version), the `mysql_upgrade` is guaranteed to fail?  It would be perfect.

10.1/root/usr/share/container-scripts/mysql/README.md
```diff
@@ -115,17 +118,65 @@ values stored in the variables and the actual passwords. Whenever a database
...
+   from the back-up.
+ * `force` -- `mysql_upgrade` is run right after the daemon has started, no matter what version the data
+   come from. This is also the way to create the missing version file `mysql_upgrade_info` if not present
+   in the root of the data directory; this file holds information about the version of the data.
+ * `optimize` -- runs `mysqlcheck --optimize` before starting the mysqld daemon, no matter what version
+   of the data is detected. It optimizes all the tables.
+ * `analyze` -- runs `mysqlcheck --analyze` before starting the mysqld daemon, no matter what version
+   of the data is detected. It analyzes all the tables.
+
+More values are separated by comma, e.g. `MYSQL_UPGRADE="optimize,analyze"`.
```
praiskup: It looks like the order mattes now, it might or might not be an issue ... if it is not, then at least a small hint should be mentioned here.

10.1/test/run
```diff
@@ -65,15 +65,21 @@ function test_connection() {
...
       echo "  Success!"
       return 0
     fi
     sleep $sleep_time
   done
-  echo "  Giving up: Failed to connect. Logs:"
+  echo " FAIL"
+  echo "  Giving up: Failed to connect."
+  echo -n "Trying to ping the IP ${ip}: "
+  ping -c 1 ${ip} &>/dev/null && echo "OK" || echo "FAIL"
```
praiskup: Meh, how it comes we have iputils installed? :)

10.1/root/usr/share/container-scripts/mysql/README.md
```diff
@@ -115,17 +118,65 @@ values stored in the variables and the actual passwords. Whenever a database
...
+   of the daemon, a warning is printed but the container starts. This is the default value.
+ * `auto` -- `mysql_upgrade` is run at the beginning of the container start, if the data version
+   can be determined and the data come with the very previous version. A warning is printed if the data
+   come from even older or newer version. This value effectively enables automatic upgrades,
+   but it is always risky and users should still back-up all the data before starting the newer container.
+   Set this option only if you have very good back-ups at any moment and you are fine to fail-over
+   from the back-up.
+ * `force` -- `mysql_upgrade` is run right after the daemon has started, no matter what version the data
+   come from. This is also the way to create the missing version file `mysql_upgrade_info` if not present
+   in the root of the data directory; this file holds information about the version of the data.
```
bparees: why is this file not always being created?

10.1/root/usr/share/container-scripts/mysql/README.md
bparees: s/More/Multiple/

10.1/root/usr/share/container-scripts/mysql/check-upgrade.sh
```diff
@@ -0,0 +1,60 @@
...
+          continue
+        fi
+
+        if [ "${datadir_version}" -eq "${mysqld_version}" ] ; then
+          log_info 'MySQL server version check OK.'
+          continue
+        fi
+
+        if [ $(( ${datadir_version} + 1 )) -eq "${mysqld_version}" -o "${datadir_version}" -eq 505 -a "${mysqld_version}" -eq 1000 ] ; then
+          log_info "MySQL server is version '${mysqld_version}' and datadir is version '${datadir_version}'."
```
bparees: this doesn't strike me as being obviously a warning.  it seems pretty easy to overlook/ignore.

10.1/root/usr/share/container-scripts/mysql/README.md
bparees: yeah explaining that the actions will be run in order is good...  most people probably want to upgrade before optimizing, i would think.

10.1/root/usr/share/container-scripts/mysql/common.sh
```diff
@@ -162,3 +169,36 @@ function wait_for_mysql_master() {
     sleep 1
   done
 }
+
+# Converts string version to the integer format (5.5.33 is converted to 505,
+# 10.1.23-MariaDB is converted into 1001, etc.
+function version2number() {
+  local version_major=${1%.*}
+  printf %d%02d ${version_major%%.*} ${version_major##*.}
+}
```
bparees: this seems fragile.  if additional dots are introduced in the string, or if the minor version number exceeds 2 digits, it will break.

10.1/root/usr/share/container-scripts/mysql/common.sh
```diff
@@ -162,3 +169,36 @@ function wait_for_mysql_master() {
...
+
+# Prints version of the mysqld that is currently available (string)
+function mysqld_version() {
+  ${MYSQL_PREFIX}/libexec/mysqld -V | awk '{print $3}'
+}
+
+# Returns version from the daemon in integer format
+function mysqld_compat_version() {
+  local compat_version=$(mysqld_version)
```
bparees: why's the point of this variable? doesn't seem used.

10.1/root/usr/share/container-scripts/mysql/README.md
hhorak: Makes sense, fixing..

10.1/test/run
hhorak: Are there systems without ping? :) Well, to be honest, what I really wanted to do is checking whether the container is running, so we should probably do it using `docker inspect...`

10.1/root/usr/share/container-scripts/mysql/README.md
hhorak: It should, but it is not. I'd call this a bug, that this PR should fix. At least I believe that all instances that will initiate the datadir after this PR is fixed, they will have the file...

10.1/root/usr/share/container-scripts/mysql/check-upgrade.sh
hhorak: Right, I'll try something less overlook-able..

10.1/root/usr/share/container-scripts/mysql/common.sh
hhorak: Agreed. Solution `version_major=$(echo "$1" | grep -o -e '^[0-9]*\.[0-9]*')` looks simplest from what I was thinking about, but if anybody has an idea how to make it correct enough without a fork, let me know..

10.1/root/usr/share/container-scripts/mysql/common.sh
hhorak: Left-over from previous implementation, I'll remove the line.

10.1/test/run
praiskup:
```
Status: Downloaded newer image for registry.access.redhat.com/rhscl/mariadb-101-rhel7:latest
bash-4.2$ ping 
bash: ping: command not found
```

10.1/root/usr/share/container-scripts/mysql/common.sh
praiskup:
```bash
#! /bin/sh -x

version2number ()
{
    set -- "${1%%.*}" "$1"
    set -- "$1" "${2##$1\.}"
    set -- "$1" "${2%%.*}"
    printf -v version2number_result %d%02d "$1" "$2"
}

version2number "$1"
echo "$version2number_result"

```

hhorak:
[test-openshift]

hhorak:
This PR is now rebased and hopefully prepared finally.

hhorak:
[test-openshift]

hhorak:
[test-openshift]

pkubatrh:
[test-openshift]

pkubatrh:
@hhorak can you update common scripts to get more waiting time for docker registry?

root-common/usr/share/container-scripts/mysql/common.sh
```diff
@@ -223,3 +228,55 @@ function process_extending_config_files() {
...
+  version2number "${version_text}"
+}
+
+# Returns name of the file in the datadir that holds version information about the data
+function get_mysql_upgrade_info_file() {
+  local datadir="$1"
+  echo "$datadir/mysql_upgrade_info"
+}
+
+# Writes version string of the daemon into mysql_upgrade_info file (should be only used when the file is missing and only when
```
kubco2: is there missing something on the end after "when"?

root-common/usr/share/container-scripts/mysql/common.sh
```diff
@@ -223,3 +228,55 @@ function process_extending_config_files() {
...
+function mysqld_version() {
+  ${MYSQL_PREFIX}/libexec/mysqld -V | awk '{print $3}'
+}
+
+# Returns version from the daemon in integer format
+function mysqld_compat_version() {
+  version2number $(mysqld_version)
+}
+
+# Returns version from the datadir
```
kubco2: Maybe add "in integer format"

root-common/usr/share/container-scripts/mysql/init/40-check-upgrade.sh
```diff
@@ -0,0 +1,111 @@
...
+                   "In order to allow seamless updates to the next higher version in the future,"\
+                   "the file mysql_upgrade_info will be created."\
+                   "If the data directory was created with a different version than ${mysqld_version_dot},"\
+                   "it is required to run this container with the MYSQL_UPGRADE environment variable"\
+                   "set to 'force', or run 'mysql_upgrade' utility manually; the mysql_upgrade tool"\
+                   "checks the tables and creates such a file as well. $(upstream_upgrade_info)"
+          write_mysql_upgrade_info_file "${MYSQL_DATADIR}"
+          continue
+          # This is currently a dead-code, but should be enabled after the mysql_upgrade_info
+          # file gets to the deployments (after few monts most of the deployments should already have the file)
```
kubco2: months

root-common/usr/share/container-scripts/mysql/init/40-check-upgrade.sh
```diff
@@ -0,0 +1,111 @@
...
+          log_warn "MySQL server is version ${mysqld_version_dot} and datadir is version"\
+                   "${datadir_version_dot}, which is a compatible combination."
+          if [ "${MYSQL_UPGRADE}" == 'auto' ] ; then
+            log_info "The data directory will be upgraded automatically from ${datadir_version_dot}"\
+                     "to version ${mysqld_version_dot}. $(upstream_upgrade_info)"
+            log_and_run mysql_upgrade ${mysql_flags}
+          else
+            log_warn "Automatic upgrade is not turned on, proceed with the upgrade."\
+                     "In order to upgrade the data directory, run this container with the MYSQL_UPGRADE"\
+                     "environment variable set to 'auto' or run running mysql_upgrade manually. $(upstream_upgrade_info)"
```
kubco2: "run running"

10.0/root/usr/share/container-scripts/mysql/README.md
```diff
@@ -260,6 +260,56 @@ Such a directory `sslapp` can then be mounted into the container with -v,
...
+This container detects whether the data needs to be upgraded using `mysql_upgrade` and
+we can control it by setting `MYSQL_UPGRADE` variable, which can have one or more of the following values:
+
+ * `warn` -- If the data version can be determined and the data come from a different version
+   of the daemon, a warning is printed but the container starts. This is the default value.
+   Since historically the version file `mysql_upgrade_info` was not created, when using this option,
+   the version file is created if not exist, but no `mysql_upgrade` will be called.
+   However, this automatic creation will be removed after few months, since the version should be
+   created on most deployments at that point.
+ * `auto` -- `mysql_upgrade` is run at the beginning of the container start, if the data version
```
kubco2: is run after container start or after daemon, like force?

root-common/usr/share/container-scripts/mysql/common.sh
hhorak: Yeah, that was supposed to say something like "only until it gets to most of the deployments"

root-common/usr/share/container-scripts/mysql/common.sh
hhorak: Thanks for the suggestion, will add this.

root-common/usr/share/container-scripts/mysql/common.sh
hhorak: Will fix..

root-common/usr/share/container-scripts/mysql/init/40-check-upgrade.sh
hhorak: hups

root-common/usr/share/container-scripts/mysql/init/40-check-upgrade.sh
hhorak: will remove 'running'

10.0/root/usr/share/container-scripts/mysql/README.md
hhorak: Both run at the same time, so will use also the same wording, good catch.

hhorak:
[test-openshift]

hhorak:
[test-openshift]

hhorak:
@kubco2, thanks for all the comments, I believe I fixed all the suggestions, and now even the tests seem to pass. Anything else you'd like to see changed before merging?

PR #35, opened 2017-05-12
Move common parts out from this project

omron93:
Common parts used for building are shared in https://github.com/sclorg/container-common-scripts and added as git submodule.

Similar PR discussed in https://github.com/sclorg/postgresql-container/pull/138

New functionality:

- allow to build also fedora based images
- allow to run OpenShift tests if implemented (https://github.com/sclorg/mongodb-container/blob/master/3.2/test/run-openshift)
- generate help.1 during the build
- sti to s2i rename

@hhorak @pkubatrh Please take a look and merge.


pkubatrh:
Tests are failing because of `help.1` now being generated in a different way than before. Otherwise LGTM. I will update the test case before merging this.
Same for https://github.com/sclorg/mysql-container/pull/183 and https://github.com/sclorg/redis-container/pull/15

omron93:
@pkubatrh Thanks. I've updated tests before reading your post :-( Thanks for help.

pkubatrh:
>@pkubatrh Thanks. I've updated tests before reading your post :-( Thanks for help.

The pipe through nroff was nice as a hotfix when we first ran into this but I really dislike this approach as nroff limits the line length of the output and introduces mid-word line breaks making grepping during tests a pain (as can actually be seen in the last test run in this PR).
Since we are now building `help.1` as part of the build process out of `README.md` I think we should just focus on testing `help.1` and not trying to use the same test on differently formatted files.

pkubatrh:
@omron93 I opened up a PR against your fork as this seems like something best to be had inside this PR: https://github.com/omron93/mariadb-container/pull/1
Please take a look if you are ok with the changes.

omron93:
@pkubatrh Thaks. It seems good. I'll prepare same fix for mysql-container and redis-container.

pkubatrh:
Tests are passing now, merging.

PR #36, opened 2017-05-23
Update Dockerfiles

omron93:
Use COPY instruction instead of ADD
(COPY is preffered for simple usage - see https://docs.docker.com/engine/userguide/eng-image/dockerfile_best-practices/#add-or-copy)

Provide same labels for all base image variants (CentOS, RHEL, Fedora)
    - add name and version labels
    - remove architecture label for RHEL based images (this label is set in RHEL image - no need to define it again)
 
(don't use separate instruction for each label)

@hhorak @praiskup @pkubatrh Please take a look and merge.

### DIFF
--- a/10.1/Dockerfile.fedora
+++ b/10.1/Dockerfile.fedora
@@ -48,7 +48,7 @@ RUN ln -s /usr/bin/python3 /usr/bin/python
 ENV CONTAINER_SCRIPTS_PATH=/usr/share/container-scripts/mysql \
     MYSQL_PREFIX=/usr
 
-ADD root /
+COPY root /
 
 # this is needed due to issues with squash
 # when this directory gets rm'd by the container-setup
--- a/10.1/Dockerfile.fedora
+++ b/10.1/Dockerfile.fedora
@@ -13,18 +13,22 @@ MAINTAINER "Honza Horak" <hhorak@redhat.com>
 #  * $MYSQL_ROOT_PASSWORD (Optional) - Password for the 'root' MySQL account
 
 ENV MYSQL_VERSION=10.1 \
-    HOME=/var/lib/mysql
+    HOME=/var/lib/mysql \
+    NAME=mariadb \
+    VERSION=10.1 \
+    RELEASE=1 \
+    ARCH=x86_64
 
 LABEL summary="MariaDB is a multi-user, multi-threaded SQL database server" \
       io.k8s.description="MariaDB is a multi-user, multi-threaded SQL database server" \
       io.k8s.display-name="MariaDB 10.1" \
       io.openshift.expose-services="3306:mysql" \
       io.openshift.tags="database,mysql,mariadb,mariadb101,galera" \
-      BZComponent="$NAME" \
-      Name="$FGC/$NAME" \
-      Version="$VERSION" \
-      Release="$RELEASE.$DISTTAG" \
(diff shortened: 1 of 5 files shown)


10.1/Dockerfile.fedora
```diff
@@ -19,14 +19,12 @@ LABEL summary="MariaDB is a multi-user, multi-threaded SQL database server" \
...
-      io.openshift.tags="database,mysql,mariadb,mariadb101,galera"
-
-ENV NAME=mariadb VERSION=10.1 RELEASE=1 ARCH=x86_64
-LABEL BZComponent="$NAME" \
-        Name="$FGC/$NAME" \
-        Version="$VERSION" \
-        Release="$RELEASE.$DISTTAG" \
-        Architecture="$ARCH"
+      io.openshift.tags="database,mysql,mariadb,mariadb101,galera" \
+      BZComponent="$NAME" \
```
hhorak: We can use the lower-case variant for Fedora dockerfile as well now (https://fedoraproject.org/wiki/Container:Guidelines#LABELS).

10.1/Dockerfile.fedora
```diff
@@ -19,14 +19,12 @@ LABEL summary="MariaDB is a multi-user, multi-threaded SQL database server" \
       io.k8s.description="MariaDB is a multi-user, multi-threaded SQL database server" \
       io.k8s.display-name="MariaDB 10.1" \
       io.openshift.expose-services="3306:mysql" \
-      io.openshift.tags="database,mysql,mariadb,mariadb101,galera"
-
-ENV NAME=mariadb VERSION=10.1 RELEASE=1 ARCH=x86_64
```
hhorak: I think removing the ENV definition will make the LABELS defined empty.. I also kind of like the idea of having environment variables like those defined, because:
* we can re-use them in the labels specification (e.g. `version="$VERSION"`)
* they will be available in the image (so, we might use them in the scripts), labels are not

So, maybe we can define them like this in RHEL/CentOS Dockerfiles as well.. The variables should be defined on separate lines though, like:
```
ENV NAME="mariadb" \
    VERSION="10.1" \
    RELEASE="1" \
    ARCH="x86_64"
```

omron93: I removed it by mistake. Thank you for pointing this out.

It is added back.
(only for Fedora now - CentOS does not use `release` and `architecture` now. Also wouldn't it cause problems during rebuilding and automatic release bumping?)

10.1/Dockerfile.fedora
omron93: Thanks, you are right. Changed.

10.1/Dockerfile.fedora
hhorak: Good point, we can check it and change later if it will work, so let's not do it for now.. I think it might still work during rebuilding, because `release` label value in upstream is ignored IMO.

10.1/Dockerfile.fedora
hhorak: Thanks.

hhorak: Actually, is it changed really? It looks like it's still not.

omron93:
> Actually, is it changed really? It looks like it's still not.

Sorry, I didn't pushed changes to remote. Now it should be OK.

PR #41, opened 2017-08-11
Do not set innodb_additional_mem_pool_size which is deprecated

hhorak:
Do not set `innodb_additional_mem_pool_size` server option which is deprecated and also not used unless `innodb_use_sys_malloc` is turned off, which is not.
See more at https://mariadb.com/kb/en/mariadb/xtradbinnodb-server-system-variables/#innodb_additional_mem_pool_size


hhorak:
No records of why this option was set: https://github.com/sclorg/mysql-container/pull/191
Merging.

PR #44, opened 2017-09-08
De-duplicating scripts

hhorak:
This change should help maintaining less files. This should be general concept usable also for other images. In principle the files that are common across versions are located in the parent directory and docker build then works with the parent directory context as well.

We basically did these changes:

    ln -sf . 10.0/10.0
    ln -sf . 10.1/10.1
    git add 10.0/10.0
    git add 10.1/10.1
    git mv 10.1/root/ root-common
    git reset HEAD -- 10.1/root/usr/share/container-scripts/mysql/README.md
    git checkout 10.1/root/usr/share/container-scripts/mysql/README.md
    git rm -f root-common/usr/share/container-scripts/mysql/README.md
    git rm -r 10.0/root/
    git reset HEAD -- 10.0/root/usr/share/container-scripts/mysql/README.md
    git checkout 10.0/root/usr/share/container-scripts/mysql/README.md
    ln -sf ../root-common 10.0/root-common
    ln -sf ../root-common 10.1/root-common
    git add 10.*/root-common


hhorak:
@FaramosCZ @pkubatrh @bkabrda FYI, no effective changes, just a bit of shuffling..

pkubatrh:
As per earlier communication let us remove the symlinks to $VERSION inside the versioned directories for now. They are not be needed for github and we should be able to work around them elsewhere.

Otherwise LGTM.

hhorak:
Yeah, true, my initial idea was to keep the current workflow still working, aka:
`cd 10.0 ; docker build .`
...but it does not obviously work anyway, so I've removed them.

pkubatrh:
Thanks, tests are passing so merging.

PR #46, opened 2017-10-04
Some files should have non-default attributes, so let's make sure they have them

hhorak:


hhorak:
This should be fixed in s2i-base now.

PR #48, opened 2017-10-04
Add templates from openshift/origin.

pvalena:
[Changes Info](https://github.com/sclorg/s2i-python-container/pull/218#issuecomment-335276731)


hhorak:
A slight race-condition, there are now examples in yaml format added already, see https://github.com/sclorg/mariadb-container/tree/master/examples
Or are these different ones?

pvalena:
Yes, I know of the redundance.
I'll just verify the conversion then(as YAML is preffered), so this PR will probably be just deleted. (I've created them all by a script.)

pvalena:
I've got one question though( @bparees, maybe you could answer me ):
Is the conversion fine, as in the openshift/origin repository you've maintained the JSON format. Is it internal OpenShift representation?

bparees:
yaml is preferred for new content, but for existing content if you rename it from .json to .yaml you're going to cause a lot of headaches.

pvalena:
@bparees, so I take it the conversion while moving to our repositories would cause headaches...
@hhorak, I'll delete your .yaml files then.

hhorak:
@bparees where the headaches come from?

bparees:
@hhorak all our logic that sucks in template files in various places.  openshift library, openshift's bindata file, test code....

pvalena:
@bparees, are we to maintain some of those other files too, bindata f.e.?
Also, are there some check we could run? Apart from actually processing the templates?

bparees:
When/if you deliver changes to origin you'll be expected to rebuild and deliver the updated bindata w/ it. You just have to run hack/update-generated-bindata.sh in origin.

Places where the existing templates reside today:

Place where a lot of users pick them up and what should be considered the canonical source of the template today (and what our tests rely on):
https://github.com/openshift/origin/tree/master/examples/db-templates

Our attempt at a centralized library (imports from openshift/origin):
https://github.com/openshift/library/tree/master/official/mariadb/templates
(https://github.com/openshift/library/blob/master/official.yaml#L133-L141)

Where our installer ships them (I think this pulls in from openshift/origin, not sure)
https://github.com/openshift/openshift-ansible/tree/master/roles/openshift_examples/files/examples/v3.7/db-templates

For application examples we have a similar model, except the canonical source is the example repos, e.g.:

https://github.com/openshift/nodejs-ex/tree/master/openshift/templates

which is imported into origin:
https://github.com/openshift/origin/tree/master/examples/quickstarts

and into library:
https://github.com/openshift/library/tree/master/official/nodejs/templates

and into ansible-install:
https://github.com/openshift/openshift-ansible/tree/master/roles/openshift_examples/files/examples/v3.7/quickstart-templates

So again, if you start changing the names of things, it will need to filter down through a lot of places.

pvalena:
PTAL.

hhorak:
@bparees, do I understand correctly that the templates must stay where they are now? If so, I'm wondering how the synchronization might work. I mean maintaining templates together with the images makes perfect sense, it allows us to potentially change templates together with a new feature, and test the templates on each PR. What is not clear to me is how the templates get to the other places -- how often it is necessary to pull them and whether it is already automated somehow in other cases..

bparees:
> @bparees, do I understand correctly that the templates must stay where they are now? 

no, we can move them anywhere we want, but whereever the "official" versions live, we still need a process to get them into the openshift/library and openshift/origin repositories.

>  I mean maintaining templates together with the images makes perfect sense, it allows us to potentially change templates together with a new feature, and test the templates on each PR

yes and no.  the templates are also somewhat tied to the example applications, which is why for the language examples (e.g. nodejs-ex) the templates live with the sample app repo, not the image repo.

> What is not clear to me is how the templates get to the other places -- how often it is necessary to pull them and whether it is already automated somehow in other cases..

openshift/library will automatically import from the locations it is told to import from.

openshift/origin does not automatically import anything, we manually update it. (there are some scripts to help with that).

Note that for the db-templates, the origin repo *IS* the canonical location today (looks like you want to change that, and that's fine, but we need to make sure we are updating origin still).
https://github.com/openshift/origin/tree/master/examples/db-templates

for the quickstarts, the canonical location today is the sample repos, hence the readme that says not to update the files directly:
https://github.com/openshift/origin/tree/master/examples/quickstarts

but rather to run this script to refresh them:
https://github.com/openshift/origin/blob/master/hack/update-external-examples.sh

Once things get into the origin and library repos, various other components consume/package them for releases, namely openshift-ansible consumes them for building the installer, and openshift online consumes them for providing them to the online environment.

The ideal world i'd like to get to is:

1) imagestreams+templates live in their canonical location (I don't care where this is)
2) openshift/library imports/aggregates them from (1)
3) openshift/origin, openshift-ansible, and openshift-online import from (2)

but there are some release process challenges around that (we don't want someone to update (1), have (2) pick it up, and that breaks something in (3), so we need finer grained control over what (3) actually picks up)

bparees:
and moving locations is less problematic than renaming things (hence the json/yaml issue).

hhorak:
[test]

PR #50, opened 2017-10-12
Keep version and relese synced with downstream

pkubatrh:


omron93:
@hhorak @pkubatrh Can't this cause some problems for already build/pushed images?

pkubatrh:
No issues for rhel based images since this is what was already used in them. For centos ones I have no idea.

But we should keep them in sync (rhel vs. centos) in any case.

omron93:
So rhel image internally is built with version 1. But centos image in Docker Hub has tag 10.2. So after several improvements in this repo, latest image will have tag 1 but it will be confusing because there would be image with tag 10.2.
So I am thinking about removing the 10.2 image from Docker Hub. @pkubatrh @hhorak Ideas?

hhorak:
I'm definitely for keeping the versions in sync, and if removing the image is possible from docker hub, let's do it. If it is not, I can live with that... Most people should use `:latest` tag anyway.

hhorak:
Thanks for the PR, merging.

omron93:
> I'm definitely for keeping the versions in sync, and if removing the image is possible from docker hub, let's do it. If it is not, I can live with that...

It is possible - remove image from docker hub. And then push image (with right tag) again. Unfortunately I don't know Hub credentials to be able to do it.

hhorak:
It is also possible to remove the tag through web UI, so it's solved now.

omron93:
Thanks.

PR #53, opened 2017-11-01
Update maintainer as a label

ficap:


ficap:
closing because maintainer label is inherited from s2i-core image

PR #54, opened 2017-11-09
add usage and description labels

TomasTomecek:
https://github.com/projectatomic/ContainerApplicationGenericLabels/blob/master/vendor/redhat/labels.md#other-labels

We're in a process of decommissioning our modular images: https://github.com/container-images/mariadb/pull/3. Since the modular variant is just about `dnf install -y $the_module` I don't think it makes sense to create a dedicated Dockerfile for such thing.

### DIFF
--- a/10.1/Dockerfile.rhel7
+++ b/10.1/Dockerfile.rhel7
@@ -12,9 +12,8 @@ FROM rhscl/s2i-core-rhel7
 
 ENV MYSQL_VERSION=10.1 \
     APP_DATA=/opt/app-root/src \
-    HOME=/var/lib/mysql
-
-ENV SUMMARY="MariaDB 10.1 SQL database server" \
+    HOME=/var/lib/mysql \
+    SUMMARY="MariaDB 10.1 SQL database server" \
     DESCRIPTION="MariaDB is a multi-user, multi-threaded SQL database server. The container \
 image provides a containerized packaging of the MariaDB mysqld daemon and client application. \
 The mysqld server daemon accepts connections from clients and provides access to content from \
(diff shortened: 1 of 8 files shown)


10.2/Dockerfile.fedora
```diff
@@ -28,7 +28,13 @@ LABEL summary="MariaDB is a multi-user, multi-threaded SQL database server" \
       version="$VERSION" \
       release="$RELEASE.$DISTTAG" \
       architecture="$ARCH" \
-      maintainer="SoftwareCollections.org <sclorg@redhat.com>"
+      maintainer="SoftwareCollections.org <sclorg@redhat.com>" \
+      usage="docker run -e MYSQL_USER=<user_name> -e MYSQL_PASSWORD=<password> -e MYSQL_DATABASE=<db_name> -e MYSQL_ROOT_PASSWORD=<root_password> -p 3306:3306 mariadb" \
```
omron93: IMHO there is no image named "mariadb". I think `$FGC/$NAME` should be used.

TomasTomecek: Very good point!

TomasTomecek:
Good feedback, thank you!

So I tried to unify summary, description and usage labels across all dockerfiles, does it make sense?

omron93:
> So I tried to unify summary, description and usage labels across all dockerfiles, does it make sense?

Sure. Good improvement.

Now, it LGTM. Thank you for this contribution.

[test]

10.1/Dockerfile.rhel7
```diff
@@ -13,8 +13,7 @@ FROM rhscl/s2i-core-rhel7
 ENV MYSQL_VERSION=10.1 \
     APP_DATA=/opt/app-root/src \
     HOME=/var/lib/mysql
```
omron93: @TomasTomecek Missing '\\'

TomasTomecek: well spotted! (I spent 10 minutes checking all backslashes -- wasn't good enough)

TomasTomecek:
updated

omron93:
[test]

omron93:
Tests are passing now. Merging.

@TomasTomecek Thanks for this fix.

TomasTomecek:
thank you for getting it through!

PR #57, opened 2017-12-05
Fix openshift test runs

pkubatrh:


omron93:
[test-openshift]

pkubatrh:
Since I have postgresql openshift tests in recent memory I started to work on these as well. There are still some issues present - images somehow get deleted during the test run and s2i test often fails, hence WIP.

pkubatrh:
[test-openshift]

pkubatrh:
[test-openshift]

pkubatrh:
[test-openshift]

pkubatrh:
[test-openshift]

pkubatrh:
For some reason the s2i builders seem to be unable to access github... Works ok on my boxes.

pkubatrh:
@omron93 @pvalena Any idea what that ^ happens?

pvalena:
Strange indeed.

@hhorak, does the OpenShift instance have internet access?

@pkubatrh, you could try running `ping 8.8.8.8` or similar in the builder pod.

test/run-openshift
```diff
@@ -104,15 +121,13 @@ function test_mysql_s2i() {
...
 
-ct_os_install_in_centos
 ct_os_cluster_up
 test_mysql_pure_image ${IMAGE_NAME}
 test_mysql_template ${IMAGE_NAME}
-test_mysql_s2i ${IMAGE_NAME} "https://github.com/hhorak/mariadb-container.git#s2i-support-3" test/test-app
-
-
+# TODO: Can we make the build against examples inside the same PR?
+test_mysql_s2i ${IMAGE_NAME} "https://github.com/sclorg/mariadb-container.git" test/test-app
```
pvalena: You could try omitting the `.git` from url.

pkubatrh: This should be working with `.git` as all of oc's help messages have it in there but I can try it out

pkubatrh:
[test-openshift]

pkubatrh:
[test-openshift]

pkubatrh:
[test-openshift]

pkubatrh:
[test-openshift]

pkubatrh:
@hhorak @pvalena ptal

pvalena:
@pkubatrh, can you please put template updates and `run-openshift` changes separately (into another PR)? So we can confirm tests work with the older version of templates at first.

omron93:
> can you please put template updates and run-openshift changes separately (into another PR)? So we can confirm tests work with the older version of templates at first.

Only change is in specifiing version of image. Without it CI tests will test three times same image. Obviously from openshift CI logs now we are testing also mariadb 10.1 image - what is original template doing.

pkubatrh:
Marek has a point there. The only change to the templates here is the addition of the MARIADB_VERSION argument, which makes it possible to use the template with other versions of the image.

omron93:
> which makes it possible to use the template with other versions of the image.

And this only change works! - from logs all versions are tested. So this change couldn't broke anything.

pvalena:
Hmm, I do see your point.
I'm generally against changing (addition is not an issue) `run-openshift` tests while editing the `templates` at the same time. IOW fitting the tests (unknowingly) to the change, which in turn needn't be valid.

I agree that, in this case the template change is simple and there's no need to test it seperately.
Especially when all versions get tested, which is awesome - I did not know that :).

Thanks!

pvalena:
I've just realized that `default value` of `version` parameter in template (and also `latest`) do not get tested. TBH the version/container gets tested, but not the template.
I think additional tests (utilizing `oc process -f ...`) ought to be checking that.
WDYT?

omron93:
> TBH the version/container gets tested, but not the template.

I thought that we will test streams in OpenShift somehow separatelly. In that case it make sense to me to test "default template settings" with this - that it works with OpenShift stream.
@pvalena What will be a workflow for stream testing?

pvalena:
`image-streams` are not connected to this AFAIK. And when they're tested, no template will be used for that.
This is the only place the template gets CI tested, until it goes to OpenShift.

Let me sync with you later on the workflow of that.

pkubatrh:
Were we to add a test case for default values we would still need to override the default namespace, otherwise we might run into issues trying to get an image from an imagestream that does not exist yet...

But makes sense to me to have such a specific test case.

pvalena:
We [ought](https://github.com/sclorg/mariadb-container/blob/master/test/run-openshift#L78) to never use the default namespace. Only the containers we build in CI should get tested- and used for tests. For every such image the `image-stream` is autocreated with [image push](https://github.com/sclorg/mariadb-container/blob/master/test/run-openshift#L77).

That's why I do not like the `image_name` variable [fallback](https://github.com/sclorg/mariadb-container/blob/master/test/run-openshift#L70). It's not what we should ever want to run in CI.

pvalena:
Also, to test just a template, no image has to exist. As I wrote earlier- you just use [oc process](https://docs.openshift.com/container-platform/latest/dev_guide/templates.html#generating-a-list-of-objects). That's a completely different kind of tests.

pkubatrh:
>Also, to test just a template, no image has to exist. As I wrote earlier- you just use oc process.

Can you elaborate on what would be worth checking via `oc process`? Or some example of such a test already used somewhere?

Possibly should also move this to another issue.

pvalena:
There are no such tests so far. That's why I was hoping to sync with you in person before creating any.

hhorak:
@pvalena I don't think we should include databases examples as a quickstart, since the database image and a template (without any extension) is enough as a quickstart. The s2i functionality in case of database daemons is something slightly different than what is s2i in case of ruby, httpd, nodejs, php... while in case of ruby, httpd, nodejs, or php, the image is nothing without an application, databases are just fine in default config. The s2i in case of databases is instead a way to provide something on top of it, for special features which database image does not offer by default. Hope this explanation makes sense.

pvalena:
It does, thanks!

PR #58, opened 2017-12-06
Drop -z parameter of readlink to avoid warning message

caringi:
Drop the following warning message:

test/run: line 31: warning: command substitution: ignored null byte in
input


omron93:
LGTM, thanks.

Tests are passing and this is a small change. So merging now.

PR #60, opened 2017-12-09
Add MYSQL_LOG_QUERIES_ENABLED variable same as in MySQL

hhorak:


hhorak:
Already reviewed in MySQL container, so tests passing should be fine here.

hhorak:
[test-openshift]

PR #63, opened 2018-01-10
templates: omit version/tag 10.0 from description

pvalena:
https://github.com/openshift/origin/pull/17922#discussion_r160709178

### DIFF
--- a/examples/mariadb-ephemeral-template.json
+++ b/examples/mariadb-ephemeral-template.json
@@ -246,7 +246,7 @@
     {
       "name": "MARIADB_VERSION",
       "displayName": "Version of MariaDB Image",
-      "description": "Version of MariaDB image to be used (10.0, 10.1, 10.2 or latest).",
+      "description": "Version of MariaDB image to be used (10.1, 10.2 or latest).",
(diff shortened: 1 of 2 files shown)


omron93:
[test] [test-openshift]

pvalena:
This is a description change and does not affect runtime in any way. And tests succeeded. Merging.

omron93:
Mariadb 10.0 is still supported. @pvalena Why to remove it from template?

pvalena:
@omron93, templates are for OpenShift, right? And Mariadb 10.0 is not in OpenShift.

See https://github.com/openshift/origin/pull/17922#discussion_r160709178.

omron93:
> @omron93, templates are for OpenShift, right? And Mariadb 10.0 is not in OpenShift.

I've read this. But the fact that "there's no mariadb:10.0 tag" don't imply that the tag shouldn't be there. 

I've found in OpenShift docs that 'rhscl/mariadb-100-rhel7' is supported! - https://docs.openshift.com/container-platform/3.7/using_images/db_images/mariadb.html

Issue #65, opened 2018-01-25
Add Galera packages to container image


jcpowermac:
It would be great if the image had installed the galera packages even if it wasn't configured.  
Then I wouldn't need to do something like this:

https://github.com/jcpowermac/mariadb-ha-poc/blob/b7df7f4ec06bac5127ae9bd46a3ea76a25f9a512/mariadb-galera-ephemeral-template.yml#L195-L204

The configuration can be handled using S2I (which is very cool btw).

PR #67, opened 2018-02-02
Adding galera packages and exposing ports

jcpowermac:
This PR adds the necessary package and exposes the ports required to run Galera.

Resolves https://github.com/sclorg/mariadb-container/issues/65

Testing was done in this repository:
https://github.com/jcpowermac/mariadb-ha-poc/blob/master/Dockerfile

### DIFF
--- a/10.2/Dockerfile
+++ b/10.2/Dockerfile
@@ -30,14 +30,14 @@ LABEL summary="$SUMMARY" \
       usage="docker run -d -e MYSQL_USER=user -e MYSQL_PASSWORD=pass -e MYSQL_DATABASE=db -p 3306:3306 centos/mariadb-102-centos7" \
       maintainer="SoftwareCollections.org <sclorg@redhat.com>"
 
-EXPOSE 3306
+EXPOSE 3306 4567 4568 4444
 
 # This image must forever use UID 27 for mysql user so our volumes are
 # safe in the future. This should *never* change, the last test is there
 # to make sure of that.
 RUN yum install -y centos-release-scl-rh && \
     yum-config-manager --enable centos-sclo-rh-testing && \
-    INSTALL_PKGS="rsync tar gettext hostname bind-utils groff-base shadow-utils rh-mariadb102" && \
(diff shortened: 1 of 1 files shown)


omron93:
Why `which` package is installed explicitly?

[test] [test-openshift]

jcpowermac:
@omron93 I was testing with this earlier maybe it was left over.  I don't need it let me remove it.  I did this PR to get a feel if its going to be accepted.  I can update the other Dockerfiles as well.

jcpowermac:
@omron93 I am wrong we need which...
```
/opt/rh/rh-mariadb102/root/usr//bin/wsrep_sst_common: line 120: which: command not found
```

hhorak:
I'm fine with doing these changes as a first step for having the galera replication ready, although I imagine that there should also be something more in the future -- scripts that would set up galera replication similarly to the master/slave replication and the only thing a user would need to do would be setting 'MYSQL_CLUSTER_NODES' (done probably automatically in k8s environment). I've already worked on this a bit previously, but did not really made it a good enough for even a PR:
https://github.com/hhorak/mariadb-container/commit/ae8ca1505a9eb40ed1690d8c87971662f1dd0cac

Anyway, there is also a package `rh-mariadb102-mariadb-server-galera`, that ships few tools (`clustercheck`, `galera_recovery`) and `galera.cnf`. However, the `galera.cnf` enables wsrep features once installed, so it could influenced the use cases where wsrep is not ready, so I agree with this variant of PR, that does not installs `rh-mariadb102-mariadb-server-galera`.

That said, @jcpowermac, do you already have some experiences locally with using this mariadb image together with the galera?

hhorak:
[test]
[test-openshift]

jcpowermac:
@hhorak hmm the tools though might be useful right?  So my use case is purely openshift/kube.  My scripts are here: https://github.com/jcpowermac/mariadb-ha-poc
Under `image-customization`.  There are incrediably simple and probably wouldn't be any use to this project.  What was wrong with yours?  Maybe it makes sense to move forward with your branch instead?

tchughesiv:
imho the branches don't appear to be conflicting... any reason both use-cases couldn't be supported?

jcpowermac:
@hhorak what can we do to get this PR moving?

PR #69, opened 2018-02-13
add imagestreams+templates to image repo

bparees:


bparees:
@pvalena ptal, strawman for tomorrow's discussion.  (thought is you could also have a development dir for content related to unreleased images)

pvalena:
Well, I was rather thinking of having all image-streams in our repo [image-streams repo](https://github.com/sclorg/image-streams/tree/staging) [moving to master later on].

The intention being that all image-streams would be (at least partially) auto-generated from respective *-container repositories (and `templates`).

IMHO it does not make much sense to me having both released and developement. We're inclined to having only `developement` ones.

@praiskup @hhorak @pkubatrh WDYT?

bparees:
> Well, I was rather thinking of having all image-streams in our repo image-streams repo [moving to master later on].

i think it makes more sense to define the imagestream in the same repo as the images, but it's really up to you.

> IMHO it does not make much sense to me having both released and developement. We're inclined to having only developement ones.

the other option would be to have a release branch of the repo itself and have the openshift/library pull from your release branch instead of the master branch.

At the end of the day, it's up to you I guess.  What I care about is that openshift/library has a place to pull the imagestreams from that contains your released/stable content.  But I am trying to get this flow locked down this sprint so it'd be good if we can get to a decision shortly so i can start updating openshift/library appropriately.

bparees:
@praiskup @pvalena @pkubatrh @hhorak bump, how do we move this forward?  I need somewhere to land these templates/imagestreams this week.

pvalena:
AFAIR `released` folder was the only way to solve this. So for me, this looks good, but let's make sure I've understood everything correctly and have some others'  confirmation.

praiskup:
I for one dislike the `released/` namespace.  I personally would rather have only  one
template in repo, and keep the compatibility with so far released images.  So unless
this is going to be enforced everywhere, I think the best would be to let this on
particular container maintainer.

pvalena:
@praiskup, right, so every [breaking] change takes time to propagate 1 release cycle? How do we ensure images' backward compatibility to *all* deployed templates?

praiskup:
> @praiskup, right, so every [breaking] change takes time to propagate 1 release cycle?

That's what I *personally* prefer, yes.  Namely, you should wait 1 cycle before you make a breaking change in **template**, but you should never make a breaking change in **image**.

> How do we ensure images' backward compatibility to all deployed templates?

There's no ultimate way, but we should test our development images against released templates.  Anyways, changes in templates are rare.

pvalena:
@bparees, so basically, these are **released** templates:

```
examples/mariadb-ephemeral-template.json
examples/mariadb-persistent-template.json
```

(In this repository.)

bparees:
@pvalena and where do you want the imagestream files?

pkubatrh:
> I personally would rather have only one
template in repo, and keep the compatibility with so far released images.

>right, so every [breaking] change takes time to propagate 1 release cycle?

After discussing this with @praiskup offline I am ok with this solution as well. Breaking changes in templates are rare enough as it is so we should not have much trouble having such templates release a few weeks after the image.

pkubatrh:
>and where do you want the imagestream files?

Putting them into the `$git_root/imagestreams` seems ok to me.

pvalena:
> Putting them into the $git_root/imagestreams seems ok to me.

I agree.

pkubatrh:
I have removed the mention of templates from the commit message since only imagestreams are added now. Otherwise LGTM, thanks. Merging

pvalena:
Thanks!

PR #70, opened 2018-02-16
Fix mariadb-ephemeral-template label

caringi:


pvalena:
Good catch!
Thanks, LGTM.

PR #75, opened 2018-04-25
Update 10.2 fedora dockerfile for f28, and run rpm-file-permissions

hhorak:


omron93:
CentOS CI infrastructure has some problems with allocating machines, that's why Fedora and CentOS tests are failing.

omron93:
Fixed by #91

PR #80, opened 2018-06-05
Make mariadb 10.1 EOL, freeze code

hhorak:
This just makes the mariadb 10.1 code frozen, so we don't loose it. Should not touch 10.2 version at all.


hhorak:
@kubco2 Could you, please, take a look whether there are no obvious issues, like forgetting something?

hhorak:
Ouch, I mixed up versions, 10.1 is obviously still supported, I wanted to freeze 10.0 instead. Closing this one and will prepare a new PR.

PR #83, opened 2018-06-15
Reverting the behvior when slave starts with initialized database

hhorak:
Fixes the same issue as: https://github.com/sclorg/mysql-container/issues/226


omron93:
Same change as in mentioned mysql PR - which seems to be working. So merging to keep images "same".

PR #84, opened 2018-08-07
adding ppc64le imagestream and examples

jeffdyoung:
Adding ppc64le based imagestream and examples for mariadb 10.2


jeffdyoung:
@hhorak I wasn't planning on s390x because OCP doesn't currently support it. We were planning to add it when (or if) that happened.

hhorak:
Merging, thanks!

PR #86, opened 2018-09-12
Fix for datadir on GlusterFS.

ferrymanders:
This will fix crashing of the MariaDB image in the OpenShift catalog on init when using GlusterFS as storage backend. 

GlusterFS will mount the volume directly on /var/lib/mysql/data and by default a new volume will have a .trashcan dir. The init scripts of MariaDB will fall over this empty dir with the error : 
`[ERROR] Invalid (old?) table or database name '.trashcan'
mysqlcheck: Got error: 1102: Incorrect database name '#mysql50#.trashcan' when selecting the database
FATAL ERROR: Upgrade failed
`

The ignore-db-dir option can be added multiple times to add on each other.

### DIFF
--- a/root-common/etc/my.cnf
+++ b/root-common/etc/my.cnf
@@ -9,4 +9,7 @@ skip_name_resolve
 # http://www.chriscalender.com/ignoring-the-lostfound-directory-in-your-datadir/
 ignore-db-dir=lost+found
 
+# GlusterFS equivalent of 'lost+found'
+ignore-db-dir=.trashcan
+
 !includedir /etc/my.cnf.d


omron93:
[test]

Sispheor:
+1 ! Please merge this :)

omron93:
Or other option could be to mount glusterfs volume to `/var/lib/mysql/` instead of `/var/lib/mysql/data`. In that case whatever FS special directories could be used. But not sure if location of mysql.sock  is then needed to be changed.

@hhorak Is `ignore-db-dir` change fine? Otherwise this PR looks good.

ferrymanders:
the mount point change is also an solution. We had this issue with the default Templates that ship with Openshift, they default mount on `/var/lib/mysql/data` , our short fix was adding --ignore-db-dir in the template, but we would like to change the least as possible on the template.

I believe that adding this to the my.cnf is the best way to go forward, this also fits the best with the already standing `ignore-db-dir=lost+found` which also wouldn't be needed if you would mount on `/var/lib/mysql` instead.

hhorak:
Sorry for the delay, merging.

PR #92, opened 2018-10-30
Add remote OpenShift tests and restructurize tests

hhorak:
The idea here is that for testing PRs, the same entrypoint (run-openshift) is used.
All the functions used for testing were moved to test-lib-mysql.sh, so some of them
might be shared between local and remote tests.

For remote tests, oc needs to be logged in to the cluster and internal templates
are used. Usually we also want to test the latest released image.


hhorak:
[test-openshift]

hhorak:
[test-openshift]

hhorak:
No negative feedback, merging.

PR #95, opened 2018-12-05
Add rhel8 version for mariadb 10.3

pkubatrh:


hhorak:
The error is actually caused by using `#!/usr/bin/python` in `/usr/bin/cgroup-limits`. I'll try to fix it in this PR, if it does not work, I'm planning to do one more PR, so will merge this one and fix it in the next one.

hhorak:
> The error is actually caused by using `#!/usr/bin/python` in `/usr/bin/cgroup-limits`. I'll try to fix it in this PR, if it does not work, I'm planning to do one more PR, so will merge this one and fix it in the next one.

And since there is a correct shebang in `/usr/bin/cgroup-limits` that is part of s2i-core, the correct fix is actually removing this duplicate file `/usr/bin/cgroup-limits` from mariadb image.

hhorak:
[test]

Issue #98, opened 2019-02-06
10.3 image not on DockerHub


hfuss:
I'm getting a 404 when following the link in the 10.3 README. Looks like its not actually being uploaded to DockerHub? Is this intentional? We'd like to use the 10.3 image without having to build it and push it to our own registry.

kubco2:
Hello, CentOS 8 is not released, it will be available on DockerHub after it will be released.

hfuss:
Looks this got released a few months ago: https://hub.docker.com/r/centos/mariadb-103-centos7/tags

PR #99, opened 2019-02-07
Fix README for mariadb 10.3

kubco2:
CentOS 8 is not released, so delete links and inform users.
Related to issue #98 .


hhorak:
Thx, merging.

PR #102, opened 2019-03-21
 Wait while shutting down server

kubco2:
https://mariadb.com/kb/en/library/mysqladmin/

### DIFF
--- a/10.1/root/usr/share/container-scripts/mysql/README.md
+++ b/10.1/root/usr/share/container-scripts/mysql/README.md
@@ -99,7 +99,7 @@ The following environment variables influence the MySQL configuration file. They
 **`MYSQL_INNODB_BUFFER_POOL_SIZE (default: 32M or 50% of available memory)`**  
        The size of the buffer pool where InnoDB caches table and index data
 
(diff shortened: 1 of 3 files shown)


hhorak:
I doubt this makes any change, I believe the admin_flags are already set properly, as they are set in the common.sh file before the `shutdown_local_mysql` function is defined at all:

https://github.com/sclorg/mariadb-container/blob/1e12e5036f488dd6968ebe3a3da6cdbac2c9edc5/root-common/usr/share/container-scripts/mysql/common.sh#L54

```
# Variables that are used to connect to local mysql during initialization
mysql_flags="-u root --socket=/tmp/mysql.sock"
admin_flags="--defaults-file=$MYSQL_DEFAULTS_FILE $mysql_flags"
```
Or do I miss why this is not used?

kubco2:
Thanks. You are right and it will not work, this script is different from mysql. I will look at the issue more.

kubco2:
Because constant sleep doesnt look like enough, I tried another idea how to fix it in https://github.com/sclorg/mysql-container/pull/253 .

PR #104, opened 2019-04-04
Do not use VOLUME command in Dockerfiles

kubco2:
As it does not work in Openshift Online for builder images for now.


kubco2:
[VOLUME definition is problematic](https://trello.com/c/m4EW0hwb/190-volume-definition-is-problematic)

PR #107, opened 2019-06-17
deprecate version 10.1

pkubatrh:


pkubatrh:
deprecate fedora version of mariadb 10.2 as well since it is based off Fedora 28 and that one is already EOL

hhorak:
LGTM

Issue #113, opened 2019-10-17
Dockerfile.rhel7 is missing for 10.3 version


phracek:
https://github.com/sclorg/mariadb-container/tree/master/10.3

There is only Dockerfile.rhel8 and Dockerfile.fedora.
Is it intended

PR #114, opened 2019-10-21
Add Dockerfile.rhel7 for 10.3 mariadb version

phracek:
MariaDB database container contains also Dockerfile.rhel7

Signed-off-by: Petr "Stone" Hracek <phracek@redhat.com>

This adds Dockerfile.rhel7 version into mariadb 10.3 version.
README.md is also updating so it reflects version 10.3.

Related issue: https://github.com/sclorg/mariadb-container/issues/113


pkubatrh:
Please add the new Dockerfile via the recommended way described in the [contribution guidelines](https://github.com/sclorg/welcome/blob/master/contribution.md#adding-a-new-version) so that we do not lose the history in latest version.

pkubatrh:
Would also be nice to have the centos7 version added in the same PR as well

phracek:
Of course, I will do it. I thought we are doing it only for new releases.

pkubatrh:
>Of course, I will do it. I thought we are doing it only for new releases.

Thanks! This is basically the same case, the difference being that the directory is already present and only the Dockerfiles are being added.

pkubatrh:
[test]

pkubatrh:
Some differences from the downstream version:

```
--- Dockerfile.rhel7	2019-10-22 13:00:09.378316660 +0200
+++ Dockerfile	2019-10-22 12:54:15.992364570 +0200
@@ -1,4 +1,4 @@
-FROM rhscl/s2i-core-rhel7
+FROM rhscl/s2i-core-rhel7:1
 
 # MariaDB image for OpenShift.
 #
@@ -39,8 +39,8 @@
 # to make sure of that.
 RUN yum install -y yum-utils && \
     prepare-yum-repositories rhel-server-rhscl-7-rpms && \
-    INSTALL_PKGS="rsync tar gettext hostname bind-utils groff-base rh-mariadb103" && \
-    yum install -y --setopt=tsflags=nodocs $INSTALL_PKGS && \
+    INSTALL_PKGS="rsync tar gettext hostname bind-utils groff-base rh-mariadb103 rh-mariadb103-mariadb-server-utils" && \
+    yum install -y --setopt=tsflags=nodocs shadow-utils $INSTALL_PKGS && \
     rpm -V $INSTALL_PKGS && \
     yum -y clean all --enablerepo='*' && \
     mkdir -p /var/lib/mysql/data && chown -R mysql.0 /var/lib/mysql && \
```

Iirc the `shadow-utils` removal is expected, but the `rh-mariadb103-mariadb-server-utils` should be installed (and I expect that is what is making the tests fail).

pkubatrh:
[test-openshift]

pkubatrh:
Centos failure on openshift is expected as the image is not yet available on dockerhub.

The rhel7 failure on openshift seems weird. Likely the cluster did not start up correctly/in-time?
In any case there will need to be a change for the default rhel7 registry, as mariadb-103-rhel7 image is only available on registry.redhat.io.
Similar change in redis: https://github.com/sclorg/redis-container/pull/53/files#diff-14dfd60f6d472d6be12d6fab73339b80L99-R101

pkubatrh:
[test-openshift]

pkubatrh:
lgtm, merging. thanks!

PR #115, opened 2020-01-15
Add s390x image stream and add 10.3 version

hhorak:
This was asked by @siamaksade and @gabemontero and consulted more that we should be ok with just one imagestream file and more tags for alternate architectures. Once merged, we'll submit a PR for https://github.com/openshift/library/blob/master/official.yaml and add `arch_ppc64le` and `arch_s390x` tags there.

@siamaksade or @gabemontero, can you take a look and verify we understood each other well and we can continue with other images this way?

### DIFF
--- a/imagestreams/mariadb-rhel7-ppc64le.json
+++ b/imagestreams/mariadb-rhel7-ppc64le.json
@@ -14,13 +14,13 @@
         "annotations": {
           "openshift.io/display-name": "MariaDB (Latest)",
           "openshift.io/provider-display-name": "Red Hat, Inc.",
-          "description": "Provides a MariaDB database on RHEL 7. For more information about using this database image, including OpenShift considerations, see https://github.com/sclorg/mariadb-container/tree/master/10.2/README.md.\n\nWARNING: By selecting this tag, your application will automatically update to use the latest version of MariaDB available on OpenShift, including major version updates.",
+          "description": "Provides a MariaDB database on RHEL 7. For more information about using this database image, including OpenShift considerations, see https://github.com/sclorg/mariadb-container/tree/master/10.3/README.md.\n\nWARNING: By selecting this tag, your application will automatically update to use the latest version of MariaDB available on OpenShift, including major version updates.",
           "iconClass": "icon-mariadb",
           "tags": "database,mariadb,ppc64le"
         },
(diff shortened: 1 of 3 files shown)


gabemontero:
Yeah I think @siamaksade tried to clarify this in the google hangout chat ... I'll try again here.

What we actually needs are separate imagestream json files for ppc and s390 in addition to the existing default one (for x86).

So undo the deletion of the ppc file you've done.

Then add a s390x one, where you have an annotation of `"tags": "database,mariadb,s390x`

then leave the tags annotation unaltered for the existing x86 one.

I'll make inline comments as well.

imagestreams/mariadb-rhel7-ppc64le.json
```diff
@@ -1,49 +0,0 @@
-{
```
gabemontero: don't delete this file ... and if the 10.3 version of mariadb is available on ppc64le, let's add that as well

Then create a `mariadb-rhel7-s390x.json` file with `"tags": "database, mariadb, s390x"` and include imagestreamtags for each version of mariadb available on s390x

imagestreams/mariadb-rhel7.json
```diff
@@ -33,7 +33,7 @@
           "openshift.io/provider-display-name": "Red Hat, Inc.",
           "description": "Provides a MariaDB 10.2 database on RHEL 7. For more information about using this database image, including OpenShift considerations, see https://github.com/sclorg/mariadb-container/tree/master/10.2/README.md.",
           "iconClass": "icon-mariadb",
-          "tags": "database,mariadb",
+          "tags": "database,mariadb,ppc64le,s390x",
```
gabemontero: nope leave this as is .... don't add ppc or s390 here

gabemontero:
in line comments added @hhorak

gabemontero:
updates look good @hhorak 

LGTM

hhorak:
@gabemontero Ok, thanks for the feedback, got it now, but will rather wait for re-check that now it is as expected.

hhorak:
@gabemontero you're super fast, reply before I manage to ask :) thanks a lot!

pkubatrh:
LGTM as well, merging!

PR #119, opened 2020-03-03
Update the image version on which this container should be built

FaramosCZ:
F29 is currently dead. F30 is currently the oldest supported Fedora version.

### DIFF
--- a/10.3/Dockerfile.fedora
+++ b/10.3/Dockerfile.fedora
@@ -1,4 +1,4 @@
-FROM registry.fedoraproject.org/f29/s2i-core:latest
+FROM registry.fedoraproject.org/f30/s2i-core:latest
 
 # MariaDB image for OpenShift.
 #


pkubatrh:
Thanks for the PR! Can you also remove the `.exclude-fedora` file (https://github.com/sclorg/mariadb-container/blob/master/10.3/.exclude-fedora) so we can try and run the tests to see what fails and what succeeds.

pkubatrh:
[test]

phracek:
@FaramosCZ The Fedora test is failing on two things:

```bash
15:39:51 /usr/bin/run-mysqld: /usr/bin/cgroup-limits: /usr/libexec/platform-python: bad interpreter: No such file or directory
15:39:51 /usr/share/container-scripts/mysql/common.sh: /usr/bin/cgroup-limits: /usr/libexec/platform-python: bad interpreter: No such file or directory
15:39:51   Success!
```
It seems that `f30/s2i-core` image does not contain fix https://github.com/sclorg/s2i-base-container/pull/198.
What about to use `f31/s2i-core` image which already contains the fix https://registry.fedoraproject.org/repo/f31/s2i-core/tags/

The second error is caused by setting key_buffer_size:

```bash
15:40:25 FAIL: option 'key_buffer_size' should have value '25M', but it wasn't found in any of the configuration files (/etc/my.cnf /etc/my.cnf.d/40-paas.cnf /etc/my.cnf.d/50-my-tuning.cnf /etc/my.cnf.d/base.cnf):
```

Can you please fix both of them? If `f31/s2i-core` can not be used, then let me know and we will build the newest `f30/s2i-core`. But I would prefer to use `f31`.

zmiklank:
This is not applicable now, closing. I will create PR with migration  f35->f37.

PR #124, opened 2020-05-22
Reorder imagestreams.

phracek:
Order is now: latest, 10.3 and 10.2.

We would like to test if the latest version is present in the imagestreams.
It is done by ansible-script (https://github.com/sclorg/ansible-tests/pull/4)

The second tag is going to be used for test if version is the latest.

Signed-off-by: Petr "Stone" Hracek <phracek@redhat.com>

### DIFF
--- a/imagestreams/mariadb-centos7.json
+++ b/imagestreams/mariadb-centos7.json
@@ -27,24 +27,6 @@
         }
       },
       {
-        "name": "10.2",
-        "annotations": {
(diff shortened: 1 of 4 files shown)


phracek:
In reference to PR, why this change is needed https://github.com/sclorg/ansible-tests/pull/4

bparees:
Seems fine to me

PR #125, opened 2020-05-27
Support MYSQL_CHARSET and MYSQL_COLLATION variables

atodorov:
like the centos/mariadb (5.5) container does

Question: I see these variables listed in multiple Dockerfile files. Should all of them be updated ?

Any ETA when this could make it into Docker Hub or a reasonable workaround ?

### DIFF
--- a/10.3/Dockerfile
+++ b/10.3/Dockerfile
@@ -9,6 +9,8 @@ FROM centos/s2i-core-centos7
 #  * $MYSQL_PASSWORD - User's password
 #  * $MYSQL_DATABASE - Name of the database to create
 #  * $MYSQL_ROOT_PASSWORD (Optional) - Password for the 'root' MySQL account
+#  * $MYSQL_CHARSET (Optional) - Default character set
+#  * $MYSQL_COLLATION (Optional) - Default collation
 
 ENV MYSQL_VERSION=10.3 \
     APP_DATA=/opt/app-root/src \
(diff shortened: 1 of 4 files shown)


hhorak:
Thanks for the contribution. I'd like to also see also tests for this, to make sure it has a desired effect. I'll take a look at that hopefully next week, unless you want to do it sooner.

For the ETA, we pull changes from this repo to the RHEL containers every ~6-8 weeks, and a new centos-based image should be available on docker hub shortly after merging.

hhorak:
Since I upgraded to Fedora 32 that has cgroups v2 by default, it makes it more complicated to test on my localhost due to https://github.com/sclorg/s2i-base-container/pull/202, so it got delayed a bit.

hhorak:
[test]

atodorov:
@hhorak thanks for adding the tests & merging. What would be the next container version to have this included ?

hhorak:
We usually pull upstream changes to RHEL and RHSCL images on ~6 weeks basis. Since one such event was quite recently, my rough guess is 4-5 weeks from now.

atodorov:
@hhorak any idea if this made it upstream or not b/c I can't tell. 

On Docker Hub I see the new env variables in the description  but when I removed my workarounds and tried using the ENV variables only I got a failure. See:
https://github.com/kiwitcms/Kiwi/pull/1817/files
https://github.com/kiwitcms/Kiwi/pull/1817/checks?check_run_id=1072417667

Issue #112, opened 2019-09-20
MariaDB 10.4 OpenShift image?


trueblue2704:
Are there any plans to have an officially supported MariaDB 10.4 image for OpenShift like you have for 10.1, 10.2?

hhorak:
Sorry, we will most probably not have officially supported 10.4 version, this one is currently planned to be skipped.

ctml91:
Will there be a 10.5 version? Stuck on 10.3 still.

hhorak:
> Will there be a 10.5 version? Stuck on 10.3 still.

We're working on it, an initial PR https://github.com/sclorg/mariadb-container/pull/145 is actually submitted already and looks good so far. Stay tuned.

phracek:
Closing this issue. We already provided mariadb-105.

PR #145, opened 2021-01-08
Add 10.5

hhorak:
This uses a new feature of "devel repos" as more described in https://github.com/sclorg/container-common-scripts/pull/186. Therefore, that PR should be merged before this one.

Except adding sources for 10.5 version, these are other changes done to make the 10.5 working:

1) From some unknown reason, listing the DBs to be replicated made the user permissions not being replicated on 10.5
It seems that filtering on mysql name was not supposed to be working anyway according to https://jira.mariadb.org/browse/MDEV-5704
And there were some changes done in 10.4 that might be related: https://jira.mariadb.org/browse/MDEV-17658
Anyway, there is probably no reason to limit the databases replicated by default.

2) Using `--auth-root-authentication-method=normal` when initiating DB, which changed the default value since 10.5 and because we don't use root account inside the container, we cannot use the new functionality that depends on being a superuser after initialization of the datadir.

Templates not updated because RPMs are not yet available, the new version is also not list in the main README.md yet.


hhorak:
[test]

hhorak:
[test]

hhorak:
[test]

hhorak:
The Fedora failure is very interesting, leads to https://bugzilla.redhat.com/show_bug.cgi?id=1914984 that is a dupe of https://bugzilla.redhat.com/show_bug.cgi?id=1908281.

hhorak:
Addresses #112.

hhorak:
[test]

phracek:
[test]

phracek:
@hhorak Thanks for the PR and driving it! Can you please update also image streams for CentOS7 at least?

10.5/cccp.yml
```diff
@@ -0,0 +1 @@
+job-id: mariadb-103-centos7
```
phracek, line 1: I guess, it should be `mariadb-105-centos7`, right?

root-common/usr/share/container-scripts/mysql/init/40-datadir-action.sh
```diff
@@ -5,6 +5,7 @@ upstream_upgrade_info() {
     10.1) echo "https://mariadb.com/kb/en/library/upgrading-from-mariadb-100-to-mariadb-101/" ;;
     10.2) echo "https://mariadb.com/kb/en/library/upgrading-from-mariadb-101-to-mariadb-102/" ;;
     10.3) echo "https://mariadb.com/kb/en/library/upgrading-from-mariadb-102-to-mariadb-103/" ;;
+    10.5) echo "https://mariadb.com/kb/en/upgrading-from-mariadb-104-to-mariadb-105/" ;;
```
phracek, line 8: We are not planning to support 10.4 in https://github.com/sclorg/mariadb-container. Therefore this is not valid, my thinking. Does it make sense to add there also upgrade from 10.3 to 10.4, so users are informed properly?

phracek:
I guess devel-repos should be mentioned in version `10.5`. Version 10.3 is already out and fully supported.

phracek:
If it is feasible now, please update also main README.md file with build reference to quay.io repository which already exists https://quay.io/repository/centos7/mariadb-105-centos7/status

hhorak:
Rebased, let's try tests again.

phracek:
Let's restart tests because of https://github.com/sclorg/rhscl-container-ci/pull/76

[test]

phracek:
@hhorak Tests are passing, but there are some issues that should be solved, I guess. Please take a look on my comments. Thanks.

phracek (changes requested):
There are some minor issues that should be solved. Please look at them.

10.5/cccp.yml
hhorak: Yes, fixing.

root-common/usr/share/container-scripts/mysql/common.sh
```diff
@@ -96,7 +96,8 @@ function initialize_database() {
   log_info 'Initializing database ...'
   log_info 'Running mysql_install_db ...'
   # Using --rpm since we need mysql_install_db behaves as in RPM
-  mysql_install_db --rpm --datadir=$MYSQL_DATADIR
+  # Using --auth-root-authentication-method=normal because we are not root in the container
+  mysql_install_db --rpm --datadir=$MYSQL_DATADIR --auth-root-authentication-method=normal
```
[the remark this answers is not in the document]
hhorak: Yes, it is available since 10.1: https://mariadb.com/kb/en/pluggable-authentication-overview/#installation-options-related-to-authentication-plugins

root-common/usr/share/container-scripts/mysql/init/40-datadir-action.sh
hhorak: It does make sense, thanks for the suggestion. Fixing.

hhorak:
> I guess devel-repos should be mentioned in version `10.5`. Version 10.3 is already out and fully supported.

Right, makes sense.

> If it is feasible now, please update also main README.md file with build reference to quay.io repository which already exists https://quay.io/repository/centos7/mariadb-105-centos7/status

Cool, will add it there.

hhorak:
[test]

hhorak:
I've created a separate PR for updating imagestreams, as we likely should merge those only after the images are released:
https://github.com/sclorg/mariadb-container/pull/150

hhorak:
@phracek Thanks for the review and suggestions. I think all of them should be addressed, with imagestream update PR prepared in https://github.com/sclorg/mariadb-container/pull/150.

phracek:
LGTM. Thanks for updating issues.

hhorak:
Merging... Tadaaaaaaaa

pkubatrh:
@hhorak I see this PR added `Dockerfile.tmp` - is this file just an oversight, or is it actually used for something?

PR #146, opened 2021-03-08
imagestream: point centos7 images to quay.io

vrutkovs:
Use quay.io as a source of the images to avoid Docker Hub ratelimiting where possible


phracek (approved):
LGTM. Let's try tests.

phracek:
[test][test-openshift]

PR #149, opened 2021-05-03
add && to allow probe to execute correctly

### LINKED ISSUE #148: Example Template Liveness/Readiness Probe Errors
The following Maria DB Templates have an incorrectly configured liveness/readiness probe:

- https://github.com/sclorg/mariadb-container/blob/master/examples/mariadb-ephemeral-template.json
- https://github.com/sclorg/mariadb-container/blob/master/examples/mariadb-persistent-template.json

When the probes are executed the following warning is logged in the pod:

```
2021-05-03 20:19:23 29 [Warning] Access denied for user 'userHFP'@'localhost' (using password: YES)
2021-05-03 20:19:27 30 [Warning] Access denied for user 'userHFP'@'localhost' (using password: YES)
2021-05-03 20:19:33 31 [Warning] Access denied for user 'userHFP'@'localhost' (using password: YES)
```

The probe should be two commands seperated by `;` or `&&`:

```json
"command": [
	"/bin/sh",
	"-i",
	"-c",
	"MYSQL_PWD=\"$MYSQL_PASSWORD\" && mysqladmin -u $MYSQL_USER ping"
]
```

Executing the command manually generates the following for the current probe:

```sh
sh-4.4$ MYSQL_PWD=\"$MYSQL_PASSWORD\" mysqladmin -u $MYSQL_USER ping
mysqladmin: connect to server at 'localhost' failed
error: 'Access denied for user 'dbuser'@'localhost' (using password: YES)'
```

With the correction:

```sh
sh-4.4$ MYSQL_PWD=\"$MYSQL_PASSWORD\" && mysqladmin -u $MYSQL_USER ping
mysqld is alive
```

strangiato:
Resolves #148

### DIFF
--- a/examples/mariadb-ephemeral-template.json
+++ b/examples/mariadb-ephemeral-template.json
@@ -116,7 +116,7 @@
                   "initialDelaySeconds": 5,
                   "exec": {
                     "command": [ "/bin/sh", "-i", "-c",
-                      "MYSQL_PWD=\"$MYSQL_PASSWORD\" mysqladmin -u $MYSQL_USER ping"]
+                      "MYSQL_PWD=\"$MYSQL_PASSWORD\" && mysqladmin -u $MYSQL_USER ping"]
                   }
                 },
                 "livenessProbe": {
@@ -124,7 +124,7 @@
                   "initialDelaySeconds": 30,
                   "exec": {
                     "command": [ "/bin/sh", "-i", "-c",
-                      "MYSQL_PWD=\"$MYSQL_PASSWORD\" mysqladmin -u $MYSQL_USER ping"]
+                      "MYSQL_PWD=\"$MYSQL_PASSWORD\" && mysqladmin -u $MYSQL_USER ping"]
                   }
                 },
                 "env": [
--- a/examples/mariadb-persistent-template.json
+++ b/examples/mariadb-persistent-template.json
@@ -133,7 +133,7 @@
                   "initialDelaySeconds": 5,
                   "exec": {
                     "command": [ "/bin/sh", "-i", "-c",
-                      "MYSQL_PWD=\"$MYSQL_PASSWORD\" mysqladmin -u $MYSQL_USER ping"]
+                      "MYSQL_PWD=\"$MYSQL_PASSWORD\" && mysqladmin -u $MYSQL_USER ping"]
                   }
                 },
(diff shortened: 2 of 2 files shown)


phracek:
Let's test Testing Farm as GitHub Action

[test-all]

examples/mariadb-ephemeral-template.json
```diff
@@ -116,15 +116,15 @@
                   "initialDelaySeconds": 5,
                   "exec": {
                     "command": [ "/bin/sh", "-i", "-c",
-                      "MYSQL_PWD=\"$MYSQL_PASSWORD\" mysqladmin -u $MYSQL_USER ping"]
+                      "MYSQL_PWD=\"$MYSQL_PASSWORD\" && mysqladmin -u $MYSQL_USER ping"]
```
phracek, line 119: I am not an expert in mariadb, but what about to use `export MYSQL_PWD=\"$MYSQL_PASSWORD\" mysqladmin -u $MYSQL_USER ping`

phracek (changes requested):

examples/mariadb-ephemeral-template.json
strangiato: Hey @phracek it has been a while since I originally raised this issue so I don't remember all of the details but the problem is that these are two commands without any delineator between them. 

The export changes the scope of the variable but it is still two commands that need some sort of delineator. 

I believe doing a command instead of && would also resolve the issue.

phracek:
[test-all]

phracek (approved):
LGTM. Thanks for fix.

examples/mariadb-ephemeral-template.json
phracek: I am sorry, but as far as I know BASH. the row 
`"MYSQL_PWD=\"$MYSQL_PASSWORD\" mysqladmin -u $MYSQL_USER ping"` means that mysqladmin consumes variable MYSQL_PWD

but
`"MYSQL_PWD=\"$MYSQL_PASSWORD\" &&  mysqladmin -u $MYSQL_USER ping"` is used for two bashes.
But may be I am wrong.

@hhorak Am I right?

phracek:
[test]

phracek:
[test-all]

PR #150, opened 2021-05-05
Add 10.5 imagestream

hhorak:
Should be merged once rh-mariadb105 and mariadb:10.5 in RHEL-8 are released.

PR #151, opened 2021-05-20
Update imagestreams for RHEL 8.4

yselkowitz:


hhorak:
Thanks!

hhorak:
[test-openshift]

hhorak:
There is already PR #150 that also adds RHEL-7 image that is now available since last week as part of RHSCL 3.7. I think we can go with that PR instead. The RHEL-8 MariaDB 10.5 part seems to be the same.

Closing in favor of #150.

PR #137, opened 2020-08-29
my.cnf: ignore netapp snapshot dir

sosiouxme:
also expand "ignore-db-dir" to "ignore-db-dirs" otherwise startup complains:
[Note] Using unique option prefix 'ignore-db-dir' is error-prone and can break in the future. Please use the full name 'ignore_db_dirs' instead.

PR #153, opened 2021-08-11
my.cnf: ignore netapp snapshot dir

phracek:
also expand ignore-db-dir to ignore-db-dirs otherwise startup complains:

[Note] Using unique option prefix 'ignore-db-dir' is error-prone and can break in the future. Please use the full name 'ignore_db_dirs' instead.

### DIFF
--- a/root-common/etc/my.cnf
+++ b/root-common/etc/my.cnf
@@ -7,9 +7,12 @@ symbolic-links = 0
 skip_name_resolve
 
 # http://www.chriscalender.com/ignoring-the-lostfound-directory-in-your-datadir/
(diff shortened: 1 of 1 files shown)


phracek:
[test][test-openshift][test-openshift-4]

phracek:
rebased PR #137

phracek:
TF RHEL-7 OpenShift 4 failure is reported here: #154 
TF CentOS-7 OpenShift 3 is reported here: https://github.com/sclorg/container-common-scripts/issues/195
centos7-openshift - OpenShift 3 cluster did not startup properly.

PR #158, opened 2021-12-14
Fix OpenShift 3 and 4 tests

phracek:
This pull request updates container-common-scripts.

And fixes OpenShift 3 and OpenShift 4 tests

Signed-off-by: Petr "Stone" Hracek <phracek@redhat.com>

PR #159, opened 2021-12-14
Support 10.5 for building on RHEL7 and CentOS7

phracek:
This pull request adds support for RHEL7 and CentOS7 for version 10.5.


phracek:
[test]

phracek:
@pkubatrh @zmiklank Please if you have time, please look at it.

pkubatrh:
[test-openshift]

pkubatrh:
The Openshift tests fails because we do not yet have the image in the quay.io registries. We only push into the registries when an image is build after a PR is merged, so we first need to merge this PR with failing Openshift test before the Openshift tests can pass.
#158 Will then check if the tests are passing when the quay.io registry has the image and follow up on any further failures encountered

PR #179, opened 2022-06-13
Use internal openshift register

phracek:
Nowadays, we use the external OpenShift registry for the testing image.
OpenShift 4 provides an internal OpenShift registry. Let's use it.


phracek:
[test-all]

phracek:
CentOS Stream 8 failed 'root' access not allowed. Let's try one more time.

[test]

zmiklank:
I see, that tests for openshift pass, and this PR LGTM. Let's merge it.

zmiklank (approved):

PR #196, opened 2023-02-27
Fix openshift4 tests

phracek:
This pull request fixes OpenShift 4 tests

### DIFF
--- a/test/test-lib-mysql.sh
+++ b/test/test-lib-mysql.sh
@@ -114,9 +114,9 @@ function test_mysql_s2i() {
 
 function test_mariadb_integration() {
   local service_name=mariadb
-  ct_os_template_exists mariadb-ephemeral && t=mariadb-ephemeral || t=mariadb-persistent
+  ct_os_template_exists mariadb-ephemeral && t=mariadb-ephemeral-template.json || t=mariadb-persistent-template.json
   ct_os_test_template_app_func "${IMAGE_NAME}" \
-                               "${t}" \
+                               "${THISDIR}/${t}" \
                                "${service_name}" \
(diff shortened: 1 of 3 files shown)


phracek:
This pull request fixes only OpenShift 4 tests. Container tests are not needed to run.

[test-openshift]

test/test-lib-mysql.sh
```diff
@@ -114,9 +114,9 @@ function test_mysql_s2i() {
 
 function test_mariadb_integration() {
   local service_name=mariadb
-  ct_os_template_exists mariadb-ephemeral && t=mariadb-ephemeral || t=mariadb-persistent
+  ct_os_template_exists mariadb-ephemeral && t=mariadb-ephemeral-template.json || t=mariadb-persistent-template.json
```
phracek, line 117: Does it make sense to split tests are run as ephemeral as persistent?

zmiklank: If I understand this correctly; if we decided to use local template `mariadb-ephemeral-template.json` we do not need to check with `ct_os_template_exists` if it already exist in the openshift do we? Then I would suppose using testing both templates.

However, haven't this test aimed at testing a template, which is already available in the openshift?

phracek: Let's test both of them.

phracek:
[test-all]

hhorak:
LGTM.

test/test-lib-mysql.sh
zmiklank: And could you please explain to me, what is the difference between using the local template and the template available in the OCP?

zmiklank: Thank you, I think I understand. In that case why do we not pull the templates from the repository but use the local ones?

pkubatrh: I expect we use local ones so we can test out changes to templates maintained in this repository as part of the PR.

zmiklank: Makes sense, thanks.

PR #197, opened 2023-03-01
tests: make sure that the testsuite ends wit error if any test fails

zmiklank:
tests: make sure that the testsuite ends wit error if any test fails

### DIFF
--- a/test/run
+++ b/test/run
@@ -25,6 +25,7 @@ run_ssl_test
 run_upgrade_test
 "
 
+
 if [ -e "${IMAGE_NAME:-}" ] ; then
   echo "Error: IMAGE_NAME must be specified"
   exit 1
@@ -76,6 +77,7 @@ function test_connection() {
   fi
   echo "Logs:"
   docker logs $(ct_get_cid $name)
+  ct_check_testcase_result 1
   return 1
 }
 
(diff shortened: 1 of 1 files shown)


zmiklank:
[test]

zmiklank:
[test]

zmiklank:
[test]

zmiklank:
[test]

phracek:
Let's re-run tests based on the merged https://github.com/sclorg/mariadb-container/pull/199

[test-all]

test/run
```diff
@@ -76,6 +77,7 @@ function test_connection() {
   fi
   echo "Logs:"
   docker logs $(ct_get_cid $name)
+  ct_check_testcase_result 1
```
phracek, line 80: This does not make sense AFAIK. docker logs are working properly in case of presence CID file.

phracek (commented):
LGTM. One nitpick. Please address my comment.

test/run
zmiklank: This is only a workaround to set TESTCATE_RESULT to 1, because I do not want to set it directly from the test function.

Please note, that the code is not
```
ct_check_testcase_result $?
```
but
```
ct_check_testcase_result 1
```

phracek (approved):
Thanks . LGTM.

Issue #198, opened 2023-03-02
MariaDB container doesn't remove anonymous users which causes the wildcard '%' NOT to evaluate to 'localhost'


ljavorsk:
From MariaDB Documentation the [Anonymous Accounts](https://mariadb.com/kb/en/create-user/#anonymous-accounts) should be removed when the database is used in production.

When this anonymous user is present, the '%' wildcard character doesn't include 'localhost' value.
This leads to the issue described in [BZ#2080804](https://bugzilla.redhat.com/show_bug.cgi?id=2080804)

This problem is also described in this [StackOverflow issue](https://serverfault.com/questions/996005/hostname-wildcard-does-not-include-localhost-or-127-0-0-0)

There are two ways to fix the problem:

1) Add 'root'@'localhost' and '$MYSQL_USER'@'localhost' users in `root-common/usr/share/container-scripts/mysql/common.sh` 
     - This solution should be used if we don't want to remove anonymous users from the database and we leave that up to the customers that use our container

2) Remove the anonymous user from the database (DELETE FROM mysql.user WHERE User='' AND Host='%'; FLUSH PRIVILEGES;) 
      - This could be an issue if any customer relies on having the anonymous users created and use them for some testing purposes
      - IMPORTANT -> If we decide to remove anonymous users, we need to remove them before creating our users, if anonymous users are not dropped before creating users on our end, the '%' wildcard won't work for 'localhost'

hhorak:
Thanks for the analysis, good work!

Option (2) looks obviously more correct, and I understand the concern about breaking somebody's setup. Therefore a few thoughts regarding the expectations about compatibility: I always take our README.md that describes usage of the container as the API we should not change unless necessary. Since we never documented anonymous users there, we might consider using them as not supported feature.

So, what I'd personally prefer is using the more correct (2) option and document this change in the README.md, stating that we're removing anonymous users for security reasons.

ljavorsk:
I've checked this issue in our RHEL/Fedora packages and it is now present there.

There is a high possibility that we do this ourselves in the init scripts (maybe some old flags or configs).

I'll look into it

ljavorsk:
Fixed

PR #200, opened 2023-03-15
Use --skip-test-db option with mysql_install_db

ljavorsk:
This option disables creation of the test database and anonymous users This fixes the issue: https://github.com/sclorg/mariadb-container/issues/198

### DIFF
--- a/root-common/usr/share/container-scripts/mysql/common.sh
+++ b/root-common/usr/share/container-scripts/mysql/common.sh
@@ -97,7 +97,8 @@ function initialize_database() {
   log_info 'Running mysql_install_db ...'
   # Using --rpm since we need mysql_install_db behaves as in RPM
   # Using --auth-root-authentication-method=normal because we are not root in the container
-  mysql_install_db --rpm --datadir=$MYSQL_DATADIR --auth-root-authentication-method=normal
+  # Using --skip-test-db because the anonymous users are not recomended in the production databases. With this option no test database and no anonymous users will be created (GH issue #198)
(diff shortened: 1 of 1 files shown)


zmiklank:
[test-all]

FaramosCZ:
Code LGTM 👍

For reference, this is how we set up the DB server in standard MariaDB RPMs in Fedora:
https://src.fedoraproject.org/rpms/mariadb/blob/rawhide/f/mariadb-prepare-db-dir.sh#_100

To be 100% sure the anonymous user won't be created, you may use
```
--skip-auth-anonymous-user
```
https://mariadb.com/kb/en/mysql_install_db/#options

FaramosCZ:
It would be nice to put the links to the upstream documentation inside of the commit message for future reference.

ljavorsk:
I've tried the `--skip-auth-anonymous-user` and the MariaDB server didn't recognize this option.

But then when I found the `--skip-test-db` option it was even better for this case because the 'test' database is also not recommended in production databases

FaramosCZ:
> I've tried the `--skip-auth-anonymous-user` and the MariaDB server didn't recognize this option.

The MariaDB server didn't recognized the option or the `mysql_install_db` script did not ?

AFAIK this option is only valid for the `mysql_install_db`, but not the server.

ljavorsk:
Yeah sorry, the `mysql_install_db` didn't

FaramosCZ:
> Yeah sorry, the `mysql_install_db` didn't

In that case it sounds like a bug worth reporting upstream

ljavorsk:
[test-all]

hhorak:
LGTM

hhorak (approved):

PR #209, opened 2023-07-13
Cleaning up the repository.

zmiklank:
[remove deprecated versions 10.0, 10.1, 10.2](https://github.com/sclorg/mariadb-container/commit/b16780c7c7a41fe8be465379587e5e525b0119dc)

[remove not used centos7 CI conf files](https://github.com/sclorg/mariadb-container/commit/67a75fa000fc32b60ae63ecbcc22ec0a4ee6de6d)

[fedora: change base image for 10.5 from deprecated f35 to f37](https://github.com/sclorg/mariadb-container/pull/209/commits/6f493ba27e052800ad883fbcd37cc2a29e1df1f7)

[disable testing and builing of 10.3 fedora image](https://github.com/sclorg/mariadb-container/pull/209/commits/1c2a3a6d1231d5cee3b99283cd94c9d762bbc8aa)


zmiklank:
[test-all]

zmiklank:
[test]

zmiklank:
I tried to find out why the weak deps of mariadb-server were not installed. But with no success.
The /etc/dnf/dnf.conf file in the Fedora 10.5 image is as follows:
```
bash-5.2$ cat /etc/dnf/dnf.conf
# see `man dnf.conf` for defaults and possible options

[main]
gpgcheck=True
installonly_limit=3
clean_requirements_on_remove=True
best=False
skip_if_unavailable=True
tsflags=nodocs
```
And also we **do not** explicitly use `--setopt=install_weak_deps=False` when installing packages.

phracek (approved):
LGTM. Thanks

zmiklank:
Let's wait for opinion on the weak deps of @FaramosCZ before merging this.

FaramosCZ:
I understood, that the previous version of the container content was built on top of `Fedora 35`: https://github.com/sclorg/mariadb-container/pull/209/commits/6f493ba27e052800ad883fbcd37cc2a29e1df1f7
And on the `Fedora 35`, the package installation worked as expected.

I've tried to use `Fedora 37`, on which the package installation worked as expected.
I've also tried `quay.io/fedora/s2i-core:latest` (which contained `Fedora 38` at the time I tested it), on which the package installation worked as expected.

I've managed to reproduce the issue only on the `quay.io/fedora/s2i-core:37`.
```
podman run -ti quay.io/fedora/s2i-core:37 bash
yum -y module enable mariadb:10.5
dnf install --setopt=tsflags=nodocs policycoreutils rsync tar gettext hostname bind-utils groff-base mariadb-server
```

The problem is twofold.
Not only the `mariadb-server-utils` package is not brought in, as it should: 
   https://src.fedoraproject.org/rpms/mariadb/blob/rawhide/f/mariadb.spec#_429
The `community-mysql` package is brought in as the mysql client, instead of `mariadb` package.

I don't know why.

---

Since the issue does not occur on the Fedora 37, it may be an issue that was present for short time during Fedora 37 life, is fixed now in Fedora 37, but it persists in the `quay.io/fedora/s2i-core:37` image.
Could you update the image and check if the issue persists ?

Or it may be other issue exclusive to the `quay.io/fedora/s2i-core:37` image.
Was is crafted correctly ?

And since on `2023-04-18` the `Fedora 38` was released, I'd suggest to use `Fedora 38 container image` instead.
The issue is not present here (please double check, though), and you get all of the bonuses of newer OS release as well.

---

Side note:
The cross-installation functionality when you can install mysql client for mariadb server and vice versa, is planned to be dropped in future by me.
It may help to avoid issues like this in the future, even though I doubt the cross-installation is the **root** cause of what we're seeing.

zmiklank:
> I've managed to reproduce the issue only on the `quay.io/fedora/s2i-core:37`.

This is interesting fact. I have checked changes in s2i-core image between migration from f35->f36->f37->f38 and they were none, except of bumping the version itself.

> Since the issue does not occur on the Fedora 37, it may be an issue that was present for short time during Fedora 37 life, is fixed now in Fedora 37, but it persists in the `quay.io/fedora/s2i-core:37` image. Could you update the image and check if the issue persists ?

It occurred to me too. The s2i-core and s2i-base are being rebuilt every Tuesday so if I remove explicit installation of mariadb-server-utils and re-run the tests they will be run on fresh image, so we will see if the thing is fixed.


> And since on `2023-04-18` the `Fedora 38` was released, I'd suggest to use `Fedora 38 container image` instead. The issue is not present here (please double check, though), and you get all of the bonuses of newer OS release as well.

Agreed. Will do that.

Thank you very much for such a insightful feedback!

zmiklank:
[test]

zmiklank:
> It occurred to me too. The s2i-core and s2i-base are being rebuilt every Tuesday so if I remove explicit installation of mariadb-server-utils and re-run the tests they will be run on fresh image, so we will see if the thing is fixed.

Actually I just realized this applies only for f38 s2i-core. In that case I would mark the s2i-core:37 as no longer supported and migrate to f38 straight away.

zmiklank:
I have rebuilt the s2i-core:f37 out of curiosity on my localhost. And the issue was no longer reproducible.

PR #211, opened 2023-08-08
disable testing and pushing to quay.io for CentOS7 image

zmiklank:
The CentOS registry has been decommissioned[1]. Alternative registries provide only old versions of CentOS base image. Because there is no known possibility currently to obtain fresh CentOS7 base container image (we do not want to update old images with yum -y update) we are stopping rebuilds of the SCLorg's s2i-images too.  Already built images will be still available on quay.io.

[1] https://lists.centos.org/pipermail/centos-devel/2023-May/142956.html

### DIFF
--- a/.github/workflows/build-and-push.yml
+++ b/.github/workflows/build-and-push.yml
@@ -14,18 +14,6 @@ jobs:
       fail-fast: false
       matrix:
         include:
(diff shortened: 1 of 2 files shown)


phracek:
[test]

phracek (approved):

PR #212, opened 2023-08-14
MariaDB 10.3 has reached EOL in June 2023

phracek:
MariaDB container in version 10.3 has reached EOL for RHEL7 in June 2023.

Let's stop test it.

### DIFF
--- a/README.md
+++ b/README.md
@@ -54,7 +54,7 @@ Choose either the CentOS7 or RHEL7 based image:
     To download it run:
 
     ```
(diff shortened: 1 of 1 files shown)


phracek:
[test]

frenzymadness:
The diff looks good to me. One, possibly unrelated, question: if we no longer support it on RHEL 7, why do we still provide it on Centos 7?

phracek:
It should be done by https://github.com/sclorg/mariadb-container/pull/211

PR #219, opened 2023-11-30
Migrate DeploymentConfig to Deployment

phracek:
Migration DeploymentConfigs to Deployment

    See more info here:
    https://docs.openshift.com/container-platform/4.12/applications/deployments/what-deployments-are.html

    Since OPenShift 4.14 DeploymentConfigs are deprecated: https://access.redhat.com/articles/7041372

It also improves the tests with officially released image.


phracek:
[test-openshift]

PR #228, opened 2024-03-25
Add proper latest tag, allow check for latest imagestreams

phracek:
Fix check for devel image. There should not be postgresql-container but mariadb-container.

Allow check for the latest imagestreams


phracek:
[test-openshift]

zmiklank (approved):

PR #234, opened 2024-04-05
Add a test for installing a plugin

hhorak:
This will fail until the plugin dir is fixed in the base.cnf file (fix suggested in #235).

PR #235, opened 2024-04-05
Fix plugin directory in the config file

hhorak:
Attempt to install the plugin currently fails with:
```
$ echo "INSTALL PLUGIN sql_errlog SONAME 'sql_errlog';" | mysql -uroot
ERROR 1126 (HY000) at line 1: Can't open shared library '/usr/lib64/mysql/plugin/sql_errlog.so' (errno: 11, cannot open shared object file: No such file or directory)
```

It's not clear to me whether this file with path spec is still needed at all, especially in versions that do not have a build for SCL (if it was added for that reason at all). Maybe this `base.cnf` might be removed, but I didn't do proper analysis to be able to say that it's possible. So, applying minimal patch meanwhile.

@FaramosCZ found this issue, thanks for that.

### DIFF
--- a/root-common/usr/share/container-scripts/mysql/pre-init/my-base.cnf.template
+++ b/root-common/usr/share/container-scripts/mysql/pre-init/my-base.cnf.template
@@ -1,5 +1,5 @@
 [mysqld]
 datadir = ${MYSQL_DATADIR}
 basedir = ${MYSQL_PREFIX}
-plugin-dir = ${MYSQL_PREFIX}/lib64/mysql/plugin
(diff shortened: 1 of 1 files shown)


hhorak:
A regression test for this plugin issue is added in #234.

hhorak:
[test]

hhorak:
[test]

hhorak:
@FaramosCZ, as you've discovered this issue, please, review and ack if the fix looks sane to you.

phracek (approved):
LGTM. But w/o background

PR #237, opened 2024-04-18
This commit fixes original file /bin/run-mysqld to /usr/bin/run-mysqld.

phracek:
This commit fixes original file /bin/run-mysqld to /usr/bin/run-mysqld.

The file /usr/bin/run-mysqld is copied from directory $VERSION/root-common where it is stored in subdirectory usr/bin/run-mysqld.

The file /bin/run-mysqld does not exist.


phracek:
Let's test it. Only container tests are needed.

[test]

PR #253, opened 2024-07-26
Update Dockerfile.fedora

hhorak:
Small changes in Fedora Dockerfile to match cXs Dockerfiles more.

### DIFF
--- a/10.11/Dockerfile.fedora
+++ b/10.11/Dockerfile.fedora
@@ -24,14 +24,14 @@ MariaDB databases on behalf of the clients."
 
 LABEL summary="$SUMMARY" \
       description="$DESCRIPTION" \
-      io.k8s.description="MariaDB is a multi-user, multi-threaded SQL database server" \
-      io.k8s.display-name="MariaDB 10.11" \
+      io.k8s.description="$DESCRIPTION" \
+      io.k8s.display-name="MariaDB $VERSION" \
       io.openshift.expose-services="3306:mysql" \
-      io.openshift.tags="database,mysql,mariadb,mariadb1011,galera" \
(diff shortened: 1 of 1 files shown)


phracek:
[test]

10.11/Dockerfile.fedora
```diff
@@ -24,14 +24,14 @@ MariaDB databases on behalf of the clients."
 
 LABEL summary="$SUMMARY" \
       description="$DESCRIPTION" \
-      io.k8s.description="MariaDB is a multi-user, multi-threaded SQL database server" \
-      io.k8s.display-name="MariaDB 10.11" \
+      io.k8s.description="$DESCRIPTION" \
+      io.k8s.display-name="MariaDB $VERSION" \
       io.openshift.expose-services="3306:mysql" \
-      io.openshift.tags="database,mysql,mariadb,mariadb1011,galera" \
+      io.openshift.tags="database,mysql,mariadb,mariadb1011,mariadb-1011" \
```
phracek, line 30: Only one nitpick.
Is it really necessary to have `mariadb-1011` and `mariadb1011`? I guess only the first is enough.

phracek (changes requested):
Only one nitpick. Otherwise LGTM.

10.11/Dockerfile.fedora
hhorak: I agree, I only added that because it's also included in cXs dockerfiles. So, we have those options:

1. have a little duplicity
2. remove the dash version from cXs
3. have inconsistency between cXs and Fedora

I think all of those is somewhat acceptable, I slightly prioritized (1.) but I'm open for any other option.

phracek: Ok, thank let's have a little duplicaty.

PR #262, opened 2025-01-14
Add support for building and testing mariadb-container on RHEL10

phracek:
The pull request is separated into two commits.

The first commit adds support for building and testing mariadb-container especially version '10.11' on RHEL10.
The second commit updates README.md file with RHEL10 support.

PR #263, opened 2025-01-16
Enhance 'chown' call

FaramosCZ:
1) Use common ':' separator instead of unusual '.' one, as disscussed here:
   https://github.com/sclorg/mariadb-container/pull/262#discussion_r1916127433

2) Use the same data type in the whole expression,
   so either only integers '27:0' or preferably only strings 'mysql:root'
   as suggested here:
   https://github.com/sclorg/mariadb-container/pull/262#discussion_r1916424440


phracek (approved):
LGTM. Thanks for the PR and update

PR #267, opened 2025-04-01
Various fixes

FaramosCZ:


FaramosCZ:
[test-all]

FaramosCZ:
[test]

FaramosCZ:
[test]

FaramosCZ:
[test]

10.11/Dockerfile.rhel10
```diff
@@ -10,54 +10,56 @@ FROM ubi10/s2i-core
 #  * $MYSQL_DATABASE - Name of the database to create
 #  * $MYSQL_ROOT_PASSWORD (Optional) - Password for the 'root' MySQL account
 
-ENV MYSQL_VERSION=10.11 \
+ENV MARIADB_VERSION=10.11 \
+    MARIADB_SHORT_VERSION=10.11 \
```
phracek, line 14: Instead of `10.11` use `1011`. Thanks

FaramosCZ:
[test]

FaramosCZ:
Hi @phracek, this MR is ready for review.

After some experimentation, I only included non-intrusive changes into this MR.
I advise doing the code review commit by commit, as that should be quite readable, unlike the squashed blob.

We discussed applying the same changes to other containerfiles too.
Once you review what I currently have, I can extend the changes to the other files you select.

As I changed the metadata (started using variables there), do we have any elegant way to diff the metadata of released image and one with these changes ?

10.11/Dockerfile.rhel10
```diff
@@ -11,53 +11,55 @@ FROM ubi10/s2i-core
...
     DESCRIPTION="MariaDB is a multi-user, multi-threaded SQL database server. The container \
 image provides a containerized packaging of the MariaDB mysqld daemon and client application. \
 The mysqld server daemon accepts connections from clients and provides access to content from \
 MariaDB databases on behalf of the clients."
 
-LABEL summary="$SUMMARY" \
-      description="$DESCRIPTION" \
-      io.k8s.description="$DESCRIPTION" \
-      io.k8s.display-name="MariaDB $VERSION" \
+LABEL summary="${SUMMARY}" \
```
phracek, line 26: Is there any reason why you switch from $SUMMARY -> ${SUMMARY} and similar.

The our container fails should meet POSIX criteria, I guess.

This change means that the rest cof the ontainers will have different naming conventions.

phracek (changes requested):
Thanks for the PR. See my comment.

10.11/Dockerfile.rhel10
phracek: But it looks like the other containers used this naming. let's have it there.

phracek (changes requested):
Can you please make this change also for `10.11/Dockerfile.rhel9`, `10.5/Dockerfile.rhel9`, and the others so we have consistent naming in the whole mariadb-container?

10.11/Dockerfile.rhel10
FaramosCZ: I used the PHP container as the example:
https://github.com/sclorg/s2i-php-container/blob/master/8.2/Dockerfile.rhel9

However even there the style is not fully coherent, so I'd suggest to unify it there.

I find this style much more readable for a human, I don't have a strong preference from the technical POV.

The readibility of this style is IMO highlighted in cases where multiple variables are used next to each other, such as:
```
io.openshift.tags="database,mysql,${NAME},${NAME}${MYSQL_SHORT_VERSION},${NAME}-${MYSQL_SHORT_VERSION}" \
```and similar, as the boundaries of the variable name are now clear.

FaramosCZ:
> Can you please make this change also for `10.11/Dockerfile.rhel9`, `10.5/Dockerfile.rhel9`, and the others so we have consistent naming in the whole mariadb-container?

Sure.

I extended the change to the all relevant 10.11 containerfiles.
- applying to: Fedora, RHEL 10, C10S, RHEL 9, C9S
- omitting: RHEL 8, C8S

Do you really wish to extend this change to the 10.5 images too?
If yes, which variants specifically?Just the RHEL 9 + C9S ones ?

-----

**!! WARNING !!**

- this introduces the ${NAME} environmental variable to the RHEL 9 and C9S variants !
   - I find this useful

- this introduces the ${VERSION} environmental variable to the RHEL 9 and C9S variants !
   - I'm not sure if I find this useful. I added it to keep the containerfiles as similar as possible.

But this is a topic for discussion.

FaramosCZ:
[test]

FaramosCZ:
I checked that the 10.5 container-files do not contain any new caveats, and extended the change to them, specifically:
- applying to 10.5 : RHEL 9, C9S

FaramosCZ:
[test]

PR #269, opened 2025-04-08
[minimization] Disable installation of weak dependencies by DNF

FaramosCZ:
otherwise 'mariadb-server' and 'mariadb-client' brings in (among others) the 'mariadb-server-utils' and 'mariadb-client-utils' which brings in the Perl stack.

If any of the sub-packages is later deemed useful to have present in the container image, we can add them case by case.

Before:
  Total download size: 31 M
  Installed size: 155 M
After:
  Total download size: 16 M
  Installed size: 106 M


FaramosCZ:
[test]

FaramosCZ:
[test-all]

FaramosCZ:
[test]

FaramosCZ:
[test]

FaramosCZ:
FTR, the tools that are in:

**mariadb-client-utils**
```
mariadb-access		Checks user access privileges based on the contents of the grant tables.
mariadb-find-rows	Extracts and prints SQL statements that match a specified condition.
mysql_find_rows		Retrieves and displays SELECT results from SQL input based on given criteria.
mysqlaccess		A diagnostic tool for checking access rights to a MariaDB/MySQL server.
```

**mariadb-server-utils**
```
mariadb-convert-table-format	Converts all tables in a database to a specified storage engine.
mariadb-dumpslow		Summarizes and sorts slow query logs to identify performance bottlenecks.
mariadb-fix-extensions		Renames table files with incorrect or outdated file name extensions.
mariadb-hotcopy			Quickly copies a database by locking tables and copying the data files.
mariadb-setpermission		Interactive script to set up user permissions easily.
mariadb-upgrade			Upgrades system tables and checks/repairs tables after a version upgrade.
mariadbd-multi			Manages multiple MariaDB server instances running on the same machine.
mysql_convert_table_format	Converts tables to a different storage engine (same as mariadb- version).
mysql_fix_extensions		Renames table files to correct file name extensions (same as mariadb- version).
mysql_setpermission		Assists in setting user privileges interactively (same as mariadb- version).
mysql_upgrade			Updates system tables and checks tables for compatibility after upgrade.
mysqld_multi			Starts, stops, or manages multiple MySQL server instances.
mysqldumpslow			Analyzes and summarizes slow query logs (same as mariadb-dumpslow).
mysqlhotcopy			Backs up MyISAM and ARCHIVE tables quickly by copying files.
perror				Displays error message descriptions for MariaDB/MySQL error codes.

```

FaramosCZ:
New changes:
- I added fixup to the previous PR that this PR is writing over
- I extended this change to C10S, C9S
- I added the tool listing from the respective \*-utils-\* subpackages to the commit message

FaramosCZ:
[test]

FaramosCZ:
[test]

FaramosCZ:
[test]

FaramosCZ:
[test]

FaramosCZ:
[test]

FaramosCZ:
[test]

FaramosCZ:
[test]

FaramosCZ:
I recalculated the space saved.

The DNF says (F42 is used in all cases):
BEFORE:
```
Installing:       130 packages
Total size of inbound packages is 54 MiB. Need to download 54 MiB.
After this operation, 231 MiB extra will be used (install 231 MiB, remove 0 B).
```
AFTER:
```
Installing:        36 packages
Total size of inbound packages is 27 MiB. Need to download 27 MiB.
After this operation, 140 MiB extra will be used (install 140 MiB, remove 0 B).
```

The Image size difference as shown by `podman images`:
```
REPOSITORY                 TAG         IMAGE ID	     CREATED        SIZE
quay.io/fedora/s2i-core    42          096f56888f6e  4 days ago     236 MB
localhost/weak-false       latest      470aeaae67df  2 minutes ago  386 MB
localhost/weak-true        latest      3900790122a2  5 minutes ago  488 MB
```
So the basic install takes around 150 MB.
The weak dependencies takes extra 100 MB on top of that.

-----

Furthermore, I moved the `mariadb-upgrade` utility from the `mariadb-server-utils` sub-package to the `mariadb-server` sub-package.

That way I kept the `mariadb-upgrade` functionality available in the container, even when the weak dependencies are not installed.

phracek (changes requested):
Thanks for bringing this pull request. The change itself looks good to me.

I have a question regarding C9S and C10S. These images are used by community.
Does it make sense to add this change into these containers?

What about version 10.5 and C9S?

Also also thank you for analyzes in description.

FaramosCZ:
> C9S and C10S ... Does it make sense to add this change into these containers?
> What about version 10.5 and C9S?

No, it does not.

I initially wanted to apply the change to C9S and C10S too. However during the development & testing, I struck the issue with the `mariadb-upgrade` utility. While it's normally only used when doing a DB upgrade between different major versions (e.g. 10.5 -> 10.11), the container logic we have implemented here relies on it being available non-stop, because of the `MYSQL_DATADIR_ACTION` env variable, through which the user can control how the `mariadb-upgrade` will be used.

Even though it is expected to be no-op in practically all events other than the DB upgrade between different major versions, user can have this `MYSQL_DATADIR_ACTION` env variable specified all the time nonetheless, which requires the utility to be present, in order to be called.

I fixed the issue in Fedora Rawhide and Fedora 42 by moving the `mariadb-upgrade` utility to the `mariadb-server` sub-package, as I described above, so it's available all the time in the container, removing this problem as a whole.

However this kind of re-packaging is not something I intend to do in RHEL 10 or 9 MariaDB RPMs, so this PR cannot be applied there.

Regarding the 10.5 version, I don't see value in maintaining it in Fedora anymore, once the 10.11 version is established as Fedora container.
The RPMs for 10.5 are based on Fedora 38 and they no longer exist in F40. Also F40 which EOL in a few weeks.
That's another hard reason why editing the 10.5 RPMs does not make sense.

-----

In conclusion, this PR intentionally only target the latest MDB in latest Fedora.
I hope these changes will be brought into the next major MDB version as well as the next RHEL (11).

phracek:
> However this kind of re-packaging is not something I intend to do in RHEL 10 or 9 MariaDB RPMs, so this PR cannot be applied there.

I think We can apply this changes to CentOS Stream 9 or CentOS Stream 10 and not RHEL9/10.

In case only Fedora is present, then after a time, we lose consistency. Fedora and CentOS Stream are kind of similar

PR #270, opened 2025-04-08
[build resilience fix] Begin doing user / group setup ourselves

FaramosCZ:
The user / group creation process develops. The UID / GID 27 is no longer guaranteed !! With the comming of sysusers.d user creation to Fedora:
  https://fedoraproject.org/wiki/Changes/RPMSuportForSystemdSysusers
packages are switching to the new technology.

That brings a wind of change into the MariaDB / MySQL packages in Fedora. Justification for the hardcoded UID / GID 27 does not exist for quite some time. MariaDB upstream use dynamic UID / GID assignment via sysusers.d for nearly a decade now.

The drop of the hardcoded UID / GID 27 in Fedora / RHEL packages may happen. We MUST stop relying on the packages to guarantee the hardcoded value at least from Fedora 42.

For that reasons:

- cherry-pick the user / group creation code from the DB package itself https://src.fedoraproject.org/rpms/mariadb10.11/blob/9cc88/f/mariadb10.11.spec#_1377 with the difference that here we want to get all of the potential error output, thus removing '>/dev/null 2>&1 || :' which is otherwise necessary in case of RPM scriptlets, as package installation MUST NOT be blocked by failed scriptlets
- put it before any package is installed
- test the UID / GID is correct both after UID / GID setup and then again after packages installation to be double sure

Expected cases:

CASE:
  We set up UID / GID ourselves, and after that DB server package tries that too during its installation via RPM scriptlet RESULT:
  The RPM scriptlet fails, as the user / group exists already, but package installas normally, resulting in correct UID / GID setup being kept CONTAINER BUILD PASS

CASE:
  We set up UID / GID ourselves, and after that DB server package tries that too during its installation via sysusers.d setup RESULT:
  sysusers.d is intentionally designed in a way that when the user / group exists already,the creation command is silently ignored, resulting in correct UID / GID setup being kept CONTAINER BUILD PASS

CASE:
  The 'mysql' user / group is *somehow* already present in the container base image already RESULT:
  the user / group creation code added by this commit fails, wich is good, as human investigation is required to find out how that possibly happened, and also we wouldn't be able to guarantee the correct UID / GID in that case. CONTAINER BUILD FAIL

### DIFF
--- a/10.11/Dockerfile.fedora
+++ b/10.11/Dockerfile.fedora
@@ -37,10 +37,12 @@ LABEL summary="${SUMMARY}" \
 
 EXPOSE 3306
 
-# This image must forever use UID 27 for mysql user so our volumes are
-# safe in the future. This should *never* change, the last test is there
-# to make sure of that.
-RUN INSTALL_PKGS="policycoreutils rsync tar gettext hostname bind-utils groff-base" && \
+# This image must forever use UID 27 for mysql user so our volumes are safe in the future. This should *never* change.
+# Instead of relying on the DB server package, we will do the setup ourselves before any package is installed
+RUN /usr/sbin/groupadd -g 27 -o -r mysql && \
+    /usr/sbin/useradd -M -N -g mysql -o -r -d ${HOME} -s /sbin/nologin -c "MySQL Server" -u 27 mysql && \
+    test "$(id mysql)" = "uid=27(mysql) gid=27(mysql) groups=27(mysql)" && \
+    INSTALL_PKGS="policycoreutils rsync tar gettext hostname bind-utils groff-base" && \
     dnf install -y --setopt=tsflags=nodocs ${INSTALL_PKGS} ${NAME}${MYSQL_VERSION}-server && \
     /usr/libexec/mysqld -V | grep -qe "${MYSQL_VERSION}\." && echo "Found VERSION ${MYSQL_VERSION}" && \
     dnf -y clean all --enablerepo='*' && \
--- a/10.11/Dockerfile.c10s
+++ b/10.11/Dockerfile.c10s
@@ -41,9 +41,8 @@ EXPOSE 3306
 # This image must forever use UID 27 for mysql user so our volumes are
 # safe in the future. This should *never* change, the last test is there
 # to make sure of that.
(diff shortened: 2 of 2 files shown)


FaramosCZ:
[test]

FaramosCZ:
[test-all]

FaramosCZ:
[test]

FaramosCZ:
[test]

phracek:
[test]

10.11/Dockerfile.fedora
```diff
@@ -37,10 +37,13 @@ LABEL summary="${SUMMARY}" \
...
-# This image must forever use UID 27 for mysql user so our volumes are
-# safe in the future. This should *never* change, the last test is there
-# to make sure of that.
-RUN INSTALL_PKGS="policycoreutils rsync tar gettext hostname bind-utils groff-base" && \
+# This image must forever use UID 27 for mysql user so our volumes are safe in the future. This should *never* change.
+# Instead of relying on the DB server package, we will do the setup ourselves before any package is installed
+RUN /usr/sbin/groupadd -g 27 -o -r mysql && \
+    /usr/sbin/useradd -M -N -g mysql -o -r -d ${HOME} -s /sbin/nologin -c "MySQL Server" -u 27 mysql && \
+    test "$(id mysql)" = "uid=27(mysql) gid=27(mysql) groups=27(mysql)" && \
+    \
```
phracek, line 45: Typo here.

FaramosCZ: Where exactly?

In the end I decided to NOT add extra `RUN` phase just for user addition, so I removed the `RUN` from line just after this one and added a blank line with just `    \` to at least visually split the blocks for human readers.

FaramosCZ:
[test]

FaramosCZ:
[test]

10.11/Dockerfile.fedora
phracek: Only visual split is not needed. The content is still readable even there is no split.

phracek (changes requested):
Please remove split.

FaramosCZ:
[test]

fila43:
LGTM

fila43 (approved):

PR #271, opened 2025-04-08
[build resilience fix] Install DB packages via versioned names

FaramosCZ:
MariaDB and MySQL package switched to the versioned package layout in Fedora and RHEL 10. The packages can be available in various states, but in all case where a specific version is wanted, specific version must be requested.

Thus container wanting MariaDB 10.11 specifically, must request to install 'mariadb10.11-server' specifically, instead of just 'mariadb-server' which will give you any version that is selected to be the distribution default the current moment.

### DIFF
--- a/10.11/Dockerfile.fedora
+++ b/10.11/Dockerfile.fedora
@@ -40,9 +40,8 @@ EXPOSE 3306
 # This image must forever use UID 27 for mysql user so our volumes are
 # safe in the future. This should *never* change, the last test is there
 # to make sure of that.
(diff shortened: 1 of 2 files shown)


FaramosCZ:
[test]

FaramosCZ:
[test]

FaramosCZ:
[test]

FaramosCZ:
[test]

FaramosCZ:
[test]

frenzymadness (approved):

PR #279, opened 2025-05-19
Deprecate the MariDB 10.5 image for Fedora

FaramosCZ:
The MariaDB 10.11 Fedora image is available.

The MariaDB 10.5 image is based on old Fedora release and the MariaDB 10.5 is not present in current Fedora releases.


FaramosCZ:
[test]

phracek (approved):
Thanks.

PR #281, opened 2025-05-28
Update container common sripts

phracek:
It updates container-common-scripts to the latest.

It removes 'auto-merge-on-demand' that run each 45 minutes, but let's call it on request.


phracek:
[test][test-openshift]

PR #282, opened 2025-06-02
Fix expression ${OS%[0-9]*} so it works with rhel10.

phracek:
Fix expression ${OS%[0-9]*} so it works with rhel10.

But in case of rhel10 it replaces to rhel1 that in templates the file rhel1.yaml does not exist.


phracek:
[test-openshift]

phracek:
Tests have passed. Let's get merge it.

PR #330, opened 2026-06-08
Add dist-gen check to GitHub action

phracek:
Add dist-gen check to GitHub action.
﻿It is needed before distgen on-boarding

For more information see https://docs.github.com/en/actions/how-tos/write-workflows/choose-what-workflows-do/use-jobs

## Summary by CodeRabbit

* **Chores**
  * Updated CI/CD workflow configuration to enhance test orchestration and reliability.

### DIFF
--- a/.github/workflows/container-tests.yml
+++ b/.github/workflows/container-tests.yml
@@ -3,10 +3,12 @@ on:
     types:
       - created
 jobs:
+  distgen-check:
+    uses: "sclorg/ci-actions/.github/workflows/distgen-check.yml@main"
   check-readme:
     uses: "sclorg/ci-actions/.github/workflows/check-readme.yml@main"
   container-tests:
-    needs: check-readme
+    needs: [check-readme,distgen-check]
     uses: "sclorg/ci-actions/.github/workflows/container-tests.yml@main"
     with:
       enabled-tests: '["container", "container-pytest", "openshift-pytest"]'


phracek:
[test]

.github/workflows/container-tests.yml
```diff
@@ -3,10 +3,12 @@ on:
     types:
       - created
 jobs:
+  distgen-check:
+    uses: "sclorg/ci-actions/.github/workflows/distgen-check.yml@main"
   check-readme:
     uses: "sclorg/ci-actions/.github/workflows/check-readme.yml@main"
   container-tests:
-    needs: check-readme
+    needs: [check-readme,distgen-check]
```
pkhartsk, line 11: Not necessary I think, but does no harm

.github/workflows/container-tests.yml
```diff
@@ -3,10 +3,12 @@ on:
     types:
       - created
 jobs:
+  distgen-check:
+    uses: "sclorg/ci-actions/.github/workflows/distgen-check.yml@main"
   check-readme:
     uses: "sclorg/ci-actions/.github/workflows/check-readme.yml@main"
```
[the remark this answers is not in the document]
pkhartsk: These workflows are under our organisation, I don't think it's necessary to version them as only we have write access

.github/workflows/container-tests.yml
phracek: What in case sources are not generated properly? The tests and builds are based on updated source. In case dist-gen failed, then we use old sources. This is mandatory

pkhartsk (approved):

.github/workflows/container-tests.yml
pkhartsk: The dist-gen check will probably never finish running after tests will have started, but I'm not against it.

